PR #15, opened 2022-09-26
Update the snsdemo repo to work with the latest sns. 

DanielThurau:
This script now works up until opening the swap as the nns-dapp was not working properly. While this isn't all the way where we want it, its further along then it was.


bitdivine:
I think this PR is now out of date, so closing.  But thank you, it helped.

PR #77, opened 2023-02-23
Bump ic commit

bitdivine:
# Motivation
The SNS canisters have advanced, the demo needs to update to stay current.

### DIFF
--- a/bin/dfx-sns-demo
+++ b/bin/dfx-sns-demo
@@ -74,8 +74,6 @@ sleep 1
 bin/dfx-neuron-prolong --network "$DFX_NETWORK"
 sleep 1
 
-./bin/dfx-sns-whitelist-me --network "$DFX_NETWORK"
-sleep 1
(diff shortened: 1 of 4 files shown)


bin/dfx-sns-demo
```diff
@@ -74,8 +74,6 @@ sleep 1
 bin/dfx-neuron-prolong --network "$DFX_NETWORK"
 sleep 1
```
bitdivine, line 76: Now whitelisting is valid for only one SNS, so instead of being run just once at setup, needs to be run for every SNS.

bin/dfx-sns-deploy
```diff
@@ -23,22 +23,10 @@ export DFX_NETWORK
   echo "    dfx nns import --network-mapping $DFX_NETWORK=local"
   exit 1
 } >&2
```
bitdivine, line 25: Stdout now has a bunch of other data in addition to the candid, which breaks parsing.  Thankfully the data we need is now saved to a JSON file, so we can get it from there instead.

PR #86, opened 2023-03-16
Fix typos, realpath, openssl static linking

dskloetd:
Some small fixes we found while trying to make snsdemo work on an M1 Mac:

1. For downloading `ic-admin` and `sns`, use a URL for openssl-static-binaries.
2. realpath on Mac doesn't work with paths that don't exist. So for /bin-other, call realpath with the directory below it instead of the directory we want to create later.
3. Remove a "Demo finished!" message in a place where the demo isn't finished yet.
4. Fix a typo where `bin/dfx-software-idl2json-install` was claiming quill was installed, rather than idl2json.
5. Fix typo "nuron" => "neuron".


bitdivine (approved):
LGTM. Thanks!

PR #96, opened 2023-03-27
Sns demo: disburse neuron and fill faucet

dskloetd:
# Motivation

To do manual and end-to-end testing the user needs access to SNS tokens.
The easiest way to get tokens is through the "Get Tokens" button, which sends tokens from a special address.

# Changes

1. Add a script to disburse the non-locked neuron after a sale is finalized.
2. Add a script to transfer tokens to the faucet (Get Tokens) address.
3. Call the scripts to fill the faucet from the sns demo.

# Testing

Ran the demo and verified that "Get Tokens" works in the UI.

### DIFF
--- a/bin/dfx-sns-demo
+++ b/bin/dfx-sns-demo
@@ -84,9 +84,12 @@ sleep 1
 
 ./bin/dfx-sns-demo-mksns --network "$DFX_NETWORK"
 sleep 1
-dfx-sns-sale-buy --network "$DFX_NETWORK"
+./bin/dfx-sns-sale-buy --network "$DFX_NETWORK"
 sleep 1
 ./bin/dfx-sns-sale-finalize --network "$DFX_NETWORK"
 sleep 1
+./bin/dfx-sns-disburse-neuron --network "$DFX_NETWORK"
+sleep 1
+./bin/dfx-sns-transfer-to-faucet --network "$DFX_NETWORK"
 
 : "Demo finished!  Hope you enjoyed the show."
--- a/bin/dfx-sns-disburse-neuron
+++ b/bin/dfx-sns-disburse-neuron
@@ -20,7 +20,7 @@ set -euo pipefail
 
 export DFX_NETWORK
 
-GOVERNANCE_CANISTER_ID="$(dfx canister call sns_root list_sns_canisters '(record {})' | idl2json | jq -r '.governance | .[]')"
+GOVERNANCE_CANISTER_ID="$(dfx canister call sns_root list_sns_canisters '(record {})' --network "$DFX_NETWORK" | idl2json | jq -r '.governance | .[]')"
 
 PRINCIPAL="$(dfx identity get-principal --network "$DFX_NETWORK")"
 
--- a/bin/dfx-sns-disburse-neuron
+++ b/bin/dfx-sns-disburse-neuron
@@ -3,6 +3,13 @@ SOURCE_DIR="$(dirname "$(realpath "${BASH_SOURCE[0]}")")"
 PATH="$SOURCE_DIR:$PATH"
 . "$SOURCE_DIR/versions.bash"
 
+print_help() {
+  cat <<-EOF
+	
+	Disburses the 1 neuron with a dissolve delay of 0.
+	EOF
+}
+
 # Source the clap.bash file ---------------------------------------------------
 source "$SOURCE_DIR/clap.bash"
 # Define options
@@ -10,6 +17,8 @@ clap.define short=n long=network desc="The dfx network to use" variable=DFX_NETW
 # Source the output file ----------------------------------------------------------
 source "$(clap.build)"
 set -euo pipefail
+
+export DFX_NETWORK
 
 GOVERNANCE_CANISTER_ID="$(dfx canister call sns_root list_sns_canisters '(record {})' | idl2json | jq -r '.governance | .[]')"
 
--- a/bin/dfx-sns-disburse-neuron
+++ b/bin/dfx-sns-disburse-neuron
@@ -1,0 +1,18 @@
+#!/usr/bin/env bash
+SOURCE_DIR="$(dirname "$(realpath "${BASH_SOURCE[0]}")")"
+PATH="$SOURCE_DIR:$PATH"
+. "$SOURCE_DIR/versions.bash"
+
+# Source the clap.bash file ---------------------------------------------------
+source "$SOURCE_DIR/clap.bash"
+# Define options
+clap.define short=n long=network desc="The dfx network to use" variable=DFX_NETWORK default="local"
+# Source the output file ----------------------------------------------------------
+source "$(clap.build)"
+set -euo pipefail
+
+GOVERNANCE_CANISTER_ID="$(dfx canister call sns_root list_sns_canisters '(record {})' | idl2json | jq -r '.governance | .[]')"
+
+NEURON_ID="$(dfx canister call "$GOVERNANCE_CANISTER_ID" list_neurons '(record {limit=10})'|idl2json|jq -r '.neurons[] | select(.dissolve_state[].DissolveDelaySeconds == "0") | .id[].id | join("; ")')"
+
+dfx canister call "$GOVERNANCE_CANISTER_ID" manage_neuron "(record { subaccount = vec { ${NEURON_ID} }; command = opt (variant { Disburse = record { } }) })"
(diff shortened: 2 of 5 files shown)


bin/dfx-sns-disburse-neuron
```diff
@@ -0,0 +1,18 @@
...
+source "$SOURCE_DIR/clap.bash"
+# Define options
+clap.define short=n long=network desc="The dfx network to use" variable=DFX_NETWORK default="local"
+# Source the output file ----------------------------------------------------------
+source "$(clap.build)"
+set -euo pipefail
+
+GOVERNANCE_CANISTER_ID="$(dfx canister call sns_root list_sns_canisters '(record {})' | idl2json | jq -r '.governance | .[]')"
+
+NEURON_ID="$(dfx canister call "$GOVERNANCE_CANISTER_ID" list_neurons '(record {limit=10})' | idl2json | jq -r '.neurons[] | select(.dissolve_state[].DissolveDelaySeconds == "0") | .id[].id | join("; ")')"
```
bitdivine, line 16: Is 10 high enough?  Can we bump this up to say 100?  I am aware that during testing I reach 10 neurons occasionally but  I'm pretty sure that I have never reached 100.

lmuntaner, line 16: `list_neurons` returns all neurons. You probably want the ones belonging to the user only.

In this case you need to pass `of_principal` parameter.

See how we do it in ic-js [here](https://github.com/dfinity/ic-js/blob/cb3e651039e6ac7524237fa338a4c6711acce479/packages/sns/src/governance.canister.ts#L83).

lmuntaner (approved):
Thanks!

Already approving, but I believe you might need to do a change.

bin/dfx-sns-disburse-neuron
```diff
@@ -0,0 +1,18 @@
+#!/usr/bin/env bash
+SOURCE_DIR="$(dirname "$(realpath "${BASH_SOURCE[0]}")")"
+PATH="$SOURCE_DIR:$PATH"
+. "$SOURCE_DIR/versions.bash"
```
bitdivine, line 4: If you define a `print_help(){}` function, you can include a description.  At present `--help` will describe the arguments but not what the script does.  For an example, please see: https://github.com/dfinity/nns-dapp/blob/main/scripts/dfx-software-ii-cache#L4

Note: The original `optparse` implementation did not  support descriptions so most of the commands don't have such a function, however I think the descriptions are helpful and would like to see all the help texts describe what the corresponding program does.

bin/dfx-sns-demo
```diff
@@ -84,9 +84,12 @@ sleep 1
 
 ./bin/dfx-sns-demo-mksns --network "$DFX_NETWORK"
 sleep 1
-dfx-sns-sale-buy --network "$DFX_NETWORK"
+./bin/dfx-sns-sale-buy --network "$DFX_NETWORK"
```
bitdivine, line 87: Note: On line 5 the bin dir is placed at the front of the path, so all these `./bin` prefixes can actually be deleted.  However it doesn't really matter either way and your change makes the commands consistent.  Nimzowitsch liked to over-protect key squares, why not over-protect running the right program?

bin/dfx-sns-transfer-to-faucet
```diff
@@ -0,0 +1,34 @@
+#!/usr/bin/env bash
+SOURCE_DIR="$(dirname "$(realpath "${BASH_SOURCE[0]}")")"
+PATH="$SOURCE_DIR:$PATH"
+. "$SOURCE_DIR/versions.bash"
+
```
bitdivine, line 5: Likewise a `print_help()` function would be nice.

bin/dfx-sns-transfer-to-faucet
```diff
@@ -0,0 +1,34 @@
...
+# Source the clap.bash file ---------------------------------------------------
+source "$SOURCE_DIR/clap.bash"
+# Define options
+clap.define short=n long=network desc="The dfx network to use" variable=DFX_NETWORK default="local"
+# Source the output file ----------------------------------------------------------
+source "$(clap.build)"
+set -euo pipefail
+
+LEDGER_CANISTER_ID="$(dfx canister call sns_root list_sns_canisters '(record {})' | idl2json | jq -r '.ledger | .[]')"
```
bitdivine, line 14: Please add `--network "${DFX_NETWORK}"` to the `dfx canister call` so that the same program can be used against testnets.

Likewise in the canister calls below and in `dfx-sns-disburse-neuron` above.

bin/dfx-sns-transfer-to-faucet
```diff
@@ -0,0 +1,34 @@
...
+from_e8s() {
+  amount_e8s="$1"
+  amount="$(awk "BEGIN {printf \"%.8f\", $amount_e8s * 0.00000001}")"
+  echo "$amount"
+}
+
+AMOUNT="$(from_e8s "$((BALANCE_E8S - FEE_E8S))")"
+FEE="$(from_e8s "$FEE_E8S")"
+
+FAUCET_ADDRESS="jg6qm-uw64t-m6ppo-oluwn-ogr5j-dc5pm-lgy2p-eh6px-hebcd-5v73i-nqe"
```
bitdivine, line 32: If I change the address here, will the faucet address change?  If not, where is the authoritative source for this information?

bitdivine (commented):
Looks good, thank you.  Some flags are needed to make sure that this works on testnets as well.

bin/dfx-sns-transfer-to-faucet
dskloetd: I was following `bin/dfx-sns-sale-buy` and it also doesn't pass --network.
I assumed that dfx would just pick up the DFX_NETWORK set by `clap` although I just realized that dfx-sns-sale-buy does `export DFX_NETWORK` where this script doesn't.

So now it does. I assume that's as good as passing --network.

bin/dfx-sns-transfer-to-faucet
dskloetd: I got it from here:
https://github.com/dfinity/nns-dapp/blob/d7574224ab58d8c4c3b6de72a72349b87e08ed77/frontend/src/lib/api/dev.api.ts#L19

I've added a comment.

bin/dfx-sns-disburse-neuron
lmuntaner: That's if you follow this process to participate in the swap.

But if you also participate manually, then you'll have more neurons, right?

bin/dfx-sns-transfer-to-faucet
bitdivine: I'm afraid that doesn't work.  buy does actually pass `--network` on lines 20, 25 and 32.  The last actually passes "local" as the network instead of `$DFX_NETWORK`.  That is a bug.

bin/dfx-sns-disburse-neuron
bitdivine: Support for `print_help()` was not in the original optparse, hence the low density of adequate help messages.

bin/dfx-sns-demo
bitdivine: It's not relevant for this PR but at some stage I would drop all the `./bin`.  This is not a blocker and not a request for change.

bin/dfx-sns-disburse-neuron
lmuntaner: Ok, so the only way to have Get Tokens work would be if we don't manually participate in the sale?

That might be worth a comment somewhere then.

bin/dfx-sns-transfer-to-faucet
dskloetd: Done. And also fixed the "local" in bin/dfx-sns-sale-buy.

bin/dfx-sns-disburse-neuron
dskloetd: Currently, the demo goes through the entire sale and finalizes it.
So if you run the demo, you can't participate in the sale.
If you run mksns manually, you can then participate and then you're on your own.
After you finalize the sale, you can manually transfer the tokens to the faucet if you want, or use the tokens from the neuron you got.
I haven't really thought about what would be useful or expected.
But the script isn't set in stone.
But it does now specify the of_principal as you suggested.

lmuntaner: Ok, thanks for the clarification.

I probably needed some context.

bitdivine:
Testing deployment on `small09`...

bitdivine:
Testing deployment on `small09`...

The test _failed_ so please hold off merging this for now.  But I haven't pored through the logs to see why yet.  Occasionally there are simple network errors.  Jumping on a train, will look as I travel.

PR #97, opened 2023-03-27
Fix hard-wired parameter

bitdivine:
# Motivation
The `dfx-sns-buy` command had network hard-wired to "local" on one line.  This works fine in CI, where the network is local, but nowhere else.

# Changes
- Pass `$DFX_NETWORK` instead

# Testing
- I have deployed to `small09` with `dfx-sns-demo --network small09`


bitdivine:
Done in another PR

PR #98, opened 2023-03-27
Clap should cope when there are no options to handle

bitdivine:
# Motivation
`clap.bash` is an argument parser, however it should still cope if no custom arguments are defined and the only arguments it has to handle are `--help` and `--verbose`.  Yet now it fails due to the use of an unset var.

# Changes
- Handle the case where no custom flags are defined.


lmuntaner (approved):
Thanks!

PR #99, opened 2023-03-27
Handle no help

bitdivine:
# Motivation
Many scripts do not yet have a help description.

# Changes
- Populate some of the help texts.  There are more to do when there is time.
- If no help text is defined, do not print an error message.

### DIFF
--- a/bin/dfx-ledger-get-icp
+++ b/bin/dfx-ledger-get-icp
@@ -3,7 +3,7 @@ set -euo pipefail
 SOURCE_DIR="$(dirname "$(realpath "${BASH_SOURCE[0]}")")"
 
 print_help() {
-	cat <<-EOF
+  cat <<-EOF
 
 	Gives a user toy ICP, taken from an affluent friar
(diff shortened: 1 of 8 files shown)


bin/dfx-ledger-get-icp
```diff
@@ -2,6 +2,15 @@
 set -euo pipefail
 SOURCE_DIR="$(dirname "$(realpath "${BASH_SOURCE[0]}")")"
 
+print_help() {
+  cat <<-EOF
+
+	Gives a user toy ICP, taken from an affluent friar
```
dskloetd, line 8: Is the below key the authority on the faucet principal? Or where is it defined? ;-)

dskloetd (approved):

bin/dfx-ledger-get-icp
bitdivine: Good point. :-) Added a reference to https://internetcomputer.org/docs/current/references/cli-reference/dfx-nns#_dfx_nns_install

PR #100, opened 2023-03-29
Put the state in the global cache

bitdivine:
# Motivation
Stat snapshots are quite large.  The state snapshot used to spin up a test environment does not normally vary by where it is used.  So, by default, a common location can be used.

# Changes
- By default, store snapshots in the global dfx cache.

### DIFF
--- a/bin/dfx-state-install
+++ b/bin/dfx-state-install
@@ -17,7 +17,7 @@ help_text() {
 # Source the clap.bash file ---------------------------------------------------
 source "$SOURCE_DIR/clap.bash"
 # Define options
-clap.define short=f long=state desc="The state file to save to" variable=DFX_STATE_FILE default="state.zip"
+clap.define short=f long=state desc="The state file to save to" variable=DFX_STATE_FILE default="$(dfx cache show)/state.zip"
 # Source the output file ----------------------------------------------------------
 source "$(clap.build)"
 
--- a/.github/workflows/run.yml
+++ b/.github/workflows/run.yml
@@ -52,7 +52,7 @@ jobs:
           ./bin/dfx-sns-demo-mksns-parallel -s 10 -m snsdemo8
       - name: Save state
         if: matrix.os != 'macos-11'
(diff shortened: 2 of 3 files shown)


bin/dfx-state-install
```diff
@@ -17,7 +17,7 @@ help_text() {
 # Source the clap.bash file ---------------------------------------------------
 source "$SOURCE_DIR/clap.bash"
 # Define options
-clap.define short=f long=state desc="The state file to save to" variable=DFX_STATE_FILE default="state.zip"
+clap.define short=f long=state desc="The state file to save to" variable=DFX_STATE_FILE default="$(dfx cache show)/state.zip"
```
dskloetd, line 20: Can we name it snsdemo-state.zip or something like that so it's clear where this came from.
Otherwise when you find this file in the cache directory it will be unclear what it is and it's also more likely to have a collision with something unrelated.
And it will also make it easier to search for the related scripts.

dskloetd, line 20: We should probably check if the file exists and give a useful error if it doesn't.

bitdivine: Yes, good idea.  Probably best done in a dedicated PR.  Small but touches several files.

bitdivine: Yes.  Actually, I think it would be best to have a search path.  Save locally, restore locally if there is a local snapshot, else get the snapshot from a global cache, on the basis that the snapshots are going to be the same for most branches.  No point in having lots of local copies, or making the user type a global path every time they want to spin up an environment.  That sort of API is called FRICTION. :-)

bitdivine: This should do the trick.  I don't have time to finish the PR now but when I do I'll ask for a review.  https://github.com/dfinity/snsdemo/pull/115/files

dskloetd:
Do you intend to finish this PR? It looks out of date.

bitdivine:
Outdated

PR #101, opened 2023-03-29
Compress state with xz

bitdivine:
# Motivation
xz compresses state files far better than zip (or gz).

# Changes
- Compress with tar.xz instead of zip.

# Tests
- Maybe a gh test, but run separately, not on every PR as it will make the slowest test even slower and will rarely change. Maybe test only when these scripts have changed?


dskloetd:
It would be safest if we save both the old .zip and the new .tar.xz until we test that we can correctly use the .tar.xz.
We probably need some kind of versioning of the state anyway. At least I can imagine other cases, where for example what the e2e test does depends on the state and we migrate from 1 setup to another.

bitdivine:
The existing state snapshot will not disappear so nothing will break so I feel that there is limited value in keeping zip files around.  They are big.  There is little technical risk in converting from zip to xz.

dskloetd (approved):
This gives you 3 days before the existing zip file expires and you have to start using the xz file before that, right?
Personally I would just create both until using the xz file but up to you.

PR #107, opened 2023-03-30
Stop and save state for mac

bitdivine:


bitdivine:
Tested.  The state snapshots built on GitHub's Intel osx runners do not work on M1 mac.  As such, there is limited value in this.

PR #108, opened 2023-03-30
Get replica state from a different directory on Mac

dskloetd:
# Motivation

Being able to save an snsdemo state once and use it multiple times is useful for Mac users as well.

# Changes

In `bin/dfx-state-save` check `uname` and use a different path to include in the state.zip depending on the OS.

# Testing

I have my own version of the `dfx-state-install` script which saves a user's existing state before replacing it with the saved state and I used that to test installing the state created by `bin/dfx-state-save`.

### DIFF
--- a/bin/dfx-state-save
+++ b/bin/dfx-state-save
@@ -18,6 +18,21 @@ clap.define short=i long=identities desc="The identities to save" variable=DFX_I
 # Source the output file ----------------------------------------------------------
 source "$(clap.build)"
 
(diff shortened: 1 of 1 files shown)


bitdivine (approved):
Approved in principle, it won't hurt, but it sounds that this won't help for M1 mac.

dskloetd:
It's still useful to make your own reusable state, even if you can't get it from GitHub.

PR #109, opened 2023-03-31
Stop before saving state

bitdivine:
# Motivation
Saving the data from a stopped replica is less likely to go wrong

### DIFF
--- a/.github/workflows/run.yml
+++ b/.github/workflows/run.yml
@@ -65,6 +65,7 @@ jobs:
           ./bin/dfx-network-stop
           sleep 1
           echo "Saving state"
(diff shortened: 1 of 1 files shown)


.github/workflows/run.yml
```diff
@@ -52,7 +52,19 @@ jobs:
           ./bin/dfx-sns-demo-mksns-parallel -s 10 -m snsdemo8
       - name: Save state
         if: matrix.os != 'macos-11'
-        run: bin/dfx-state-save --verbose
+        run: |
+          # How can you tell that I don't trust dfx stop
```
dskloetd, line 56: Can you change this to 
```
# Stop the replica to let it persist all its state before we save the state.
```

PR #110, opened 2023-03-31
Add a script to create a dummy anchor

dskloetd:
# Motivation

Going through the steps of creating an anchor and confirming the backup device and seed phrase gets tedious and slows down e2e tests.

# Changes

1. `bin/dfx-ii-create-dummy-anchor` script which adds a dummy identity anchor.
2. Call the script at the end of `bin/dfx-snsdemo`.
3. `bin/dfx-nns-import` which makes necessary change to dfx.json to be able to call internet_identity and exports declarations/internet_identity.did
4. Call `bin/dfx-nns-import` from `bin/dfx-snsdemo` right after it calls `dfx nns import`.

# Testing

Ran the script manually and confirmed that I could sign in by using existing anchor 10000.


bin/dfx-sns-demo
```diff
@@ -91,5 +91,7 @@ sleep 1
 ./bin/dfx-sns-disburse-neuron --network "$DFX_NETWORK"
 sleep 1
 ./bin/dfx-sns-transfer-to-faucet --network "$DFX_NETWORK"
+sleep 1
+./bin/dfx-ii-create-dummy-anchor --network "$DFX_NETWORK"
```
dskloetd, line 95: I wasn't sure if this should to here or only in the workflow after executing the demo.

dfx.json
```diff
@@ -1,21 +1,17 @@
...
-      "source": [
-        "src/smiley_dapp_assets/assets",
-        "dist/smiley_dapp_assets/"
-      ],
-      "type": "assets"
-    },
-    "smiley_dapp": {
-      "main": "src/smiley_dapp/main.mo",
-      "type": "motoko"
+  "internet_identity": {
```
dskloetd: dfx refused to call the canister without this, even if I specify the ID and the candid file manually.
See https://dfinity.slack.com/archives/CGA566TPV/p1680196767594109

bin/dfx-ii-create-dummy-anchor
```diff
@@ -0,0 +1,67 @@
...
+
+echo "Creating anchor."
+ANCHOR="$(dfx canister call internet_identity register "(record{pubkey=blob \"${PUB_KEY_BLOB}\";alias=\"dfx test key\";purpose=variant{authentication};key_type=variant{unknown};protection=variant{unprotected};},record{key=\"${CHALLENGE_KEY}\";chars=\"a\"})" --candid "$II_DID_FILE" --network "$DFX_NETWORK" | idl2json | jq -r .registered.user_number)"
+
+II_CANISTER_ID="$(dfx canister id internet_identity --network "$DFX_NETWORK")"
+
+# This matches the dummy key pair generated by Internet Identity here:
+# https://github.com/dfinity/internet-identity/blob/1ce860dd0601312f6521d433c4e273215a1abc5d/src/frontend/src/utils/iiConnection.ts#L57
+DUMMY_PUB_KEY_BLOB='0*0\05\06\03+ep\03!\00;j\27\bc\ce\b6\a4-b\a3\a8\d0*o\0dse2\15w\1d\e2C\a6:\c0H\a1\8bY\da)'
+DUMMY_CREDENTIAL_ID_BLOB='\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00'
```
frederikrothenberger, line 49: The credential ID is `opt` and could be omitted.

bin/dfx-ii-create-dummy-anchor
```diff
@@ -0,0 +1,67 @@
...
+# This is arbitrary as long as it's different from the public key above.
+# We won't actually use the seed phrase. We just want to skip the warning that
+# we don't have a key phrase.
+SEED_PHRASE_PUB_KEY_BLOB='00000000000000000000000000000000000000000000'
+
+add_device() {
+  DEVICE_PUB_KEY_BLOB="$1"
+  DEVICE_KEY_TYPE="$2"
+  DEVICE_PURPOSE="$3"
+  set dfx canister call internet_identity add "(${ANCHOR}, record { alias = \"\"; origin = opt \"http://${II_CANISTER_ID}.localhost:8080\"; protection = variant { unprotected }; pubkey = blob \"${DEVICE_PUB_KEY_BLOB}\"; key_type = variant { ${DEVICE_KEY_TYPE} }; purpose = variant { ${DEVICE_PURPOSE} }; credential_id = opt blob \"${DUMMY_CREDENTIAL_ID_BLOB}\"; })" --candid "$II_DID_FILE" --network "$DFX_NETWORK"
```
frederikrothenberger, line 60: dito for the  `origin`

bin/dfx-ii-create-dummy-anchor
dskloetd: Removed.

bin/dfx-ii-create-dummy-anchor
dskloetd: Done. Thanks!

dfx.json
dskloetd: Unfortunately snsdemo overwrites dfx.json so calling the script from the demo fails. Should I change the source of overwriting the dfx.json, or add the section from the script after overwriting?

declarations/internet_identity.did
bitdivine, line 1: Can you add a TODO comment to get this pushed upstream, please?  `dfx nns import` _should_ provide this but doesn't.  As it is, if the II deployed by `dfx nns install` changes, this API may be out of sync.  Out of date did files can be very confusing.

There is an emerging standard for getting did files directly from canisters.  The method name isn't stable yet.  Maybe II provides the API in such a way?  I think gthat fixing upstream is a better solution but this might be an ioption for staying in sync in the meantime.  Example of getting the `sns_aggrtegator` interface:

```
$ dfx canister call sns_aggregator interface --network mainnet --query | idl2json | jq -r
type CanisterStatusResultV2 = record {
  controller : principal;
  status : CanisterStatusType;
  freezing_threshold : nat;
  balance : vec record { vec nat8; nat };
  memory_size : nat;
  cycles : nat;
  settings : DefiniteCanisterSettingsArgs;
  idle_cycles_burned_per_day : nat;
  module_hash : opt vec nat8;
};
type CanisterStatusType = variant { stopped; stopping; running };
type Config = record { update_interval_ms : nat64; fast_interval_ms : nat64 };
type DefiniteCanisterSettingsArgs = record {
  controller : principal;
  freezing_threshold : nat;
  controllers : vec principal;
  memory_allocation : nat;
  compute_allocation : nat;
};
service : (opt Config) -> {
  get_canister_status : () -> (CanisterStatusResultV2);
  health_check : () -> (text) query;
  http_request : () -> () query;
  stable_data : () -> (text) query;
  tail_log : (opt nat16) -> (text) query;
}
```

bitdivine (approved):
Looks good in principle;

PR #113, opened 2023-04-04
Escape default

bitdivine:
# Motivation
By the way, it seems like clap doesn't allow \< or \> or even spaces in default values.

Context, I wanted the default to be "dfx-state-backeup-\<generated timestamp\>". I was then going to check if the default was changed or not and if not, I would generate a new value. I thought it would look good in --help like that.

# Changes
* Escape the default variable

# Tests
* A test has been added that exercises a default value with `< >` and similar characters.  It fails without the fix, works with the fix.

### DIFF
--- a/bin/clap.bash
+++ b/bin/clap.bash
@@ -75,7 +75,7 @@ function clap.define() {
 		clap_flag_match="${clap_flag_match}#NL#TB#TB${long}${short:+|${short}})#NL#TB#TB#TB${variable}=(); for ((i=0; i<nargs; i++)); do ${variable}+=( \"\$1\" ); shift 1; done;;"
 	fi
 	if [ "${default:-}" != "" ]; then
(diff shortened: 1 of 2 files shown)


bin/clap.bash
```diff
@@ -75,7 +75,7 @@ function clap.define() {
 		clap_flag_match="${clap_flag_match}#NL#TB#TB${long}${short:+|${short}})#NL#TB#TB#TB${variable}=(); for ((i=0; i<nargs; i++)); do ${variable}+=( \"\$1\" ); shift 1; done;;"
 	fi
 	if [ "${default:-}" != "" ]; then
-		clap_defaults="${clap_defaults}#NL${variable}=${default}"
+		clap_defaults="${clap_defaults}#NL${variable}=${default@Q}"
```
dskloetd, line 78: I have no idea what the existing code here does, but I'm approving based on the change and the tests.

dskloetd (approved):

PR #114, opened 2023-04-05
Use different replica data dir when saving a snapshot on Mac

dskloetd:
# Motivation

`bin/dfx-snapshot-save` checks the OS to determine the local replica data dir, but then doesn't do anything with it.

# Changes

1. Actually use the local replica data dir when archiving the snapshot
2. Small unrelated fix: Don't fail in `dfx-nns-import` if the `declarations` directory already exists.

# Tested

Created a snapshot on Mac and used it to run locally with changes from https://github.com/dfinity/nns-dapp/pull/2252

### DIFF
--- a/bin/dfx-nns-import
+++ b/bin/dfx-nns-import
@@ -26,7 +26,7 @@ EOF
 ) >dfx.new
 mv dfx.new dfx.json
 
-mkdir declarations
+mkdir -p declarations
 
 cat <<'EOF' >declarations/internet_identity.did
 type UserNumber = nat64;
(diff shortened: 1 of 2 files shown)


bitdivine (approved):
That was silly.  Thanks for the fix.  Does this mean that you can use CI builds locally now?  Or was this bug introduced after the compatibility check?

dskloetd:
I'm not sure what you mean by compatibility check. I run the script manually to generate a snapshot for myself. I don't use the snapshot from CI.

PR #115, opened 2023-04-06
Snapshot search path

bitdivine:
# Motivation
Restoring snapshots has two common use cases:
- User wants to restore a snapshot they just created
- User wants to restore a default snapshot
It would be nice if restoring looked in the locations corresponding to those two situations, in that order.


bitdivine:
Outdated

PR #116, opened 2023-04-11
Fix bin/dfx-network-provider on Mac

dskloetd:
# Motivation

Being able to create the same state on local development machine as on CI.

# Changes

1. Use `lsof` instead of `/proc` to determine in which directory a process was started.
2. Set `HOME` to the user's home directory before running `dfx info replica-port`. (It is set to something else in `become_user()` in `bin/dfx-sns-demo-mksns-parallel`

# Testing

1. Ran `bin/dfx-sns-demo-mksns-parallel` on my M1 mac.
2. Tested `lsof` and `eval echo "~$USER"` separately on my personal Ubuntu laptop.

### DIFF
--- a/bin/dfx-network-provider
+++ b/bin/dfx-network-provider
@@ -20,6 +20,9 @@ source "$(clap.build)"
 
 export DFX_NETWORK
 
+# Work-around required to be able to run `dfx info`.
+# A proper fix is in dfx 0.14.0.
+# TODO: Remove this once we use dfx 0.14.0.
 get_home() {
   eval echo "~$USER"
 }
--- a/bin/dfx-network-provider
+++ b/bin/dfx-network-provider
@@ -20,13 +20,19 @@ source "$(clap.build)"
 
 export DFX_NETWORK
 
+get_home() {
+  eval echo "~$USER"
+}
+
 case "${FORMAT}" in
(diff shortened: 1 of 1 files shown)


bin/dfx-network-provider
```diff
@@ -20,13 +20,19 @@ source "$(clap.build)"
 
 export DFX_NETWORK
 
+get_home() {
+  eval echo "~$USER"
+}
```
bitdivine, line 25: This gives an incorrect value if `dfx start` is run with HOME set to a different value from the one assumed by bash.  If there is an issue with HOME escaping a context to which it should be constrained, that problem should be fixed at source.

bitdivine (dismissed):

bin/dfx-network-provider
dskloetd: I've added a TODO to remove this later.

dskloetd:
@lmuntaner do you think you can have a look?
@bitdivine has been too busy to respond to my messages.
I'm happy to change it later if he prefers a different solution but I'd really like snsdemo to work for everybody now.
Happy to explain how it works.

lmuntaner (approved):
If it's temporary, I guess it's ok to merge it

PR #117, opened 2023-04-12
snsdemo: Init Internet Identity Salt

dskloetd:
# Motivation

Internet-identity needs to do some initiailization to access randomness.
By default this is done the first time a user signs in.
But this breaks if it's done multiple times simultaneously.
We want to run end-to-end tests on CI in parallel but that would currently fail for this reason.

# Changes

1. Add a script to initialize the internet-identity salt.
2. Call the script from `bin/dfx-sns-demo`

# Tested

Ran e2e tests in parallel from a clean state with and without initializing the salt and noticed that it fails without and passes with.

### DIFF
--- a/bin/dfx-ii-init-salt
+++ b/bin/dfx-ii-init-salt
@@ -1,0 +1,24 @@
+#!/usr/bin/env bash
+
+set -euo pipefail
+
+SOURCE_DIR="$(dirname "$(realpath "${BASH_SOURCE[0]}")")"
+
+print_help() {
+  cat <<-EOF
+
+  Initialize the randomness in Internet Identity.
+  Otherwise this will be done the first time a user registers and this will fail
+  if done multiple times simultaneously.
+	EOF
+}
+
+# Source the clap.bash file ---------------------------------------------------
+source "$SOURCE_DIR/clap.bash"
+# Define options
+clap.define short=n long=network desc="The dfx network to use" variable=DFX_NETWORK default="local"
+# Source the output file ----------------------------------------------------------
+source "$(clap.build)"
+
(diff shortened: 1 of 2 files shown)


dskloetd:
This actually doesn't work because nns-dapp CI reinstalls internet identity.

dskloetd:
Now that we plan to remove installing internet-identity from nns-dapp CI, I think it makes sense to have this.

bin/dfx-ii-init-salt
```diff
@@ -0,0 +1,24 @@
...
+
+SOURCE_DIR="$(dirname "$(realpath "${BASH_SOURCE[0]}")")"
+
+print_help() {
+  cat <<-EOF
+
+  Initialize the randomness in Internet Identity.
+  Otherwise this will be done the first time a user signs in and this will fail
+  if done multiple times simultaneously.
+	EOF
```
bitdivine, line 13: The herefile should be indented with tabs. The `<<-EOF` will remove the tab indentation.  As is, the actual text printout will be indented with two spaces.  See: https://tldp.org/LDP/abs/html/here-docs.html#:~:text=The%20%2D%20option%20to%20mark%20a%20here%20document%20limit%20string%20(%3C%3C%2DLimitString)%20suppresses%20leading%20tabs%20(but%20not%20spaces)%20in%20the%20output.%20This%20may%20be%20useful%20in%20making%20a%20script%20more%20readable.

bitdivine (approved):

bin/dfx-ii-init-salt
dskloetd: Done

PR #119, opened 2023-04-14
Ckbtc

bitdivine:
# Motivation
We wish to add ckbtc canisters to state snapshots, for use in e2e testing.  The first step in this process is to get the relevant wasm and did files.

# Changes
- Import the ckbtc download script from https://github.com/dfinity/nns-dapp/blob/main/scripts/ckbtc/download-ckbtc
- Flesh out the script so that it matches `dfx nns import`
- Add tests

# Tests
- The script self-checks after every installation, to catch cases where e.g. downloads fail.
- There is a fil eof tests to exercise corner cases, and this is run in CI.


bin/dfx-ckbtc-import
```diff
@@ -0,0 +1,131 @@
...
+  }
+  pushd "$CANDID_DIR"
+  get_did rs/bitcoin/ckbtc/minter/ckbtc_minter.did "${LOCAL_PREFIX}minter.did"
+  get_did rs/rosetta-api/icrc1/ledger/icrc1.did "${LOCAL_PREFIX}ledger.did"
+  get_did rs/rosetta-api/icrc1/index/index.did "${LOCAL_PREFIX}index.did"
+  popd
+
+  : Update dfx.json
+  LOCAL_PREFIX="${LOCAL_PREFIX:-}" \
+    jq ' .canisters["\(env.LOCAL_PREFIX)ledger"] = { build: "true", candid: "candid/\(env.LOCAL_PREFIX)ledger.did", wasm: "wasms/\(env.LOCAL_PREFIX)ledger.wasm", type: "custom", remote: { id: {mainnet: "mxzaz-hqaaa-aaaar-qaada-cai" } }}
```
lmuntaner, line 75: Does this change dfx.json?

lmuntaner (commented):
One question

bin/dfx-ckbtc-import
bitdivine: Yes, it adds canister entries to dfx.json, like `dfx nns install` but unlike the download script.  Is that OK for your use cases?

dskloetd (approved):

bin/dfx-ckbtc-import
lmuntaner: I'd rather have them in dfx.json directly. We already have three entries. We either use them or remove them.

bitdivine: Ok.  I'll make the addition optional, then.  This way the code is consistent with `dfx nns install` but doesn't change dfx.json if that is unwanted.

lmuntaner: thanks

PR #120, opened 2023-04-14
snsdemo: Install internet-identity when deploying local

dskloetd:
# Motivation

On NNS dapp we currently install internet-identity on CI.
We want as much as possible to be in the snapshot we take from snsdemo.
If we install internet-identity in snsdemo, we can remove the installation step from CI in nns-dapp

# Changes

1. Allow specifying the "flavor" of internet-identity wasm to `bin/dfx-software-internet-identity-ci-wasm-url`
2. Install the dev wasm of internet-identity in `bin/dfx-network-deploy` when deploying to network `local`

# Tested

Ran `bin/dfx-sns-demo` and verified that I could navigate to `http://qhbym-qaaaa-aaaaa-aaafq-cai.localhost:8080/` and get version `release-2023-04-12`


bitdivine (approved):
LGTM. Thanks!

PR #121, opened 2023-04-14
Ckbtc install

bitdivine:
# Motivation
We would like to have e2e tests for ckbtc functionality in the nns-dapp repo.

# Changes
- Add a CKBTC deployment script
  - This is the script from the nns-dapp repo, with variables such as network taken from command line flags.
- Include ckbtc canisters in the stock snapshot

# Tests
* The ckbtc deployment contains a self-test.  Additionally, there is a set of tests for the self-test itself.


bin/dfx-snapshot-stock-make
```diff
@@ -20,6 +20,12 @@ source "$(clap.build)"
 : Set up SNS state and create one finalized SNS
 dfx-sns-demo
 
+: Set up ckbtc canisters
+dfx-ckbtc-import --prefix ckbtc_
+dfx-ckbtc-import --prefix ckbtc_test_
+dfx-ckbtc-deploy --prefix ckbtc_
+dfx-ckbtc-deploy --prefix ckbtc_test_
+
```
bitdivine, line 28: @lmuntaner - The [ckbtc deployment script in the nns-dapp repo](https://github.com/dfinity/nns-dapp/blob/main/scripts/ckbtc/deploy-ckbtc) deploys with the prefix `ckbtc_test_` but dfx.json has the prefix `ckbtc_` (no test).  Which would you like, or both?

bin/dfx-ckbtc-deploy
bitdivine, line 1: This file is imported from nns-dapp with the following changes:
- Add the bash cli parser and `--help` message.
- Make the prefix a parameter.

bin/dfx-ckbtc-deploy
```diff
@@ -0,0 +1,94 @@
...
+       kyt_fee = opt 13_333;
+       kyt_principal = null;
+   }
+})" --mode=reinstall
+
+echo "Step 3: deploying ledger canister..."
+PRINCIPAL="$(dfx identity get-principal)"
+dfx deploy "${LOCAL}ledger" --network "$DFX_NETWORK" --argument "(variant {
+  Init = record {
+     token_symbol = \"ckBTC\";
```
bitdivine, line 59: Should any of this be variable?  Do you need both test ckbtc and main ckbtc and if so, should they have different symbols and token names?

bin/dfx-snapshot-stock-make
bitdivine, line 1: Here the ckbtc canisters are added to the stock snapshot, so that they are available for nns-dapp e2e tests.

Are there any exit criteria that need to be tested?  I can check that the canisters have been created.  Is anything else needed to classify a ckbtc deployment as fit for nns-dapp testing?

bin/dfx-snapshot-stock-make
lmuntaner: no `test`. I thought I removed that to be honest.

Thanks for doing this.

DOn't you need to add them to dfx.json?

bitdivine: `dfx-ckbtc-import` adds entries to dfx.json, if they are not already there.  If there are already entries, they are used.

bin/dfx-canister-check-wasm-hash
bitdivine, line 1: This is imported from the nns-dapp repo.  It probably belongs here, along with the other polyfills, anyway.

bin/dfx-ckbtc-deploy
lmuntaner: @peterpeterparker what do you think?

Is one enough for testnet?

lmuntaner (commented):
LGTM, but I'll wait for @peterpeterparker 's approval also

bin/dfx-ckbtc-deploy
peterpeterparker: Ideally our test environment would deploy ckTESTBTC and - or connect to - a Bitcoin testnet.

ckBTC without the Bitcoin testnet is just another ICRC token as any Sns token.

bitdivine: Sounds good.  My impression is that deploying ckTESTBTC should be possible but would mean more code, so it would be better to do that part in a separate PR.  If we aim to deploy ckTESTBTC, is the data in this PR compatible?  E.g. is the token name fine?  We can always adjust later but if it's right from the start we don't risk breaking e2e tests that use the first token name.

If the code here is consistent with the goal, can we merge please?

lmuntaner (approved):

bin/dfx-ckbtc-deploy
lmuntaner: The effort to have `ckTESTBTC` is big. Even the FI team don't have it. Maybe we should wait until they do?

We have a lot of things in our plates at the moment.

PR #122, opened 2023-04-14
Test release

bitdivine:


bin/dfx-snapshot-mk-stock
```diff
@@ -0,0 +1,38 @@
...
+source "$SOURCE_DIR/clap.bash"
+# Define options
+clap.define short=s long=snapshot desc="The file to save to" variable=DFX_SNAPSHOT_FILE default="stock-snsdemo-snapshot.tar.xz"
+# Source the output file ----------------------------------------------------------
+source "$(clap.build)"
+
+: Set up SNS state and create one finalized SNS
+bin/dfx-sns-demo
+
+: Stop the replica to let it persist all its state before we save the state.
```
bitdivine, line 23: Note: This logic is probably always wanted before dfx-snapshot-save so can be moved there.

bin/dfx-snapshot-mk-stock
```diff
@@ -0,0 +1,38 @@
...
+dfx stop
```
dskloetd, line 24: FYI, we'll probably want to sleep about 30 seconds before this line unless we have some way to be sure that all the requests have been handled. Because if you look in the [logs](https://github.com/dfinity/nns-dapp/actions/runs/4700088373/jobs/8334327519) of nns-dapp CI it has stuff like:
```
[Canister rrkah-fqaaa-aaaaa-aaaaq-cai] Execution of proposal: 18 succeeded.
[Canister u7xn3-ciaaa-aaaaa-aaa4a-cai] Refresh_buyer_tokens for buyer 7rleq-s52u4-wganc-do3c5-6kbms-3k5wp-gao6x-kk5oy-jkvaq-7vjt5-aqe; old e8s 0; new e8s 314100000000
[Canister u7xn3-ciaaa-aaaaa-aaa4a-cai] Swap has reached ICP target of 314100000000
[Canister u7xn3-ciaaa-aaaaa-aaa4a-cai] Token swap committed; 1 direct investors and 0 community fund investors receive a total of 314100000000 out of 314100000000 (change 0);
```
which doesn't really make sense to have when using a stock snapshot.

.github/workflows/snapshot.yml
```diff
@@ -0,0 +1,52 @@
...
+            : Optionally, make the release public
+            [[ "$tag" != release-* ]] || {
+                    echo "Making release public"
+                    gh release edit "$tag" --draft=false
+            }
+          done
+          : Updates 'tip' release, if this was triggered by a push to main or 'test-release'.
+          : This can be exercised without a PR by pushing to 'test-release'.
+          if [[ "${{github.event_name}}" == 'push' ]] && ( [[ "${{github.ref_name}}" == 'main' ]] || [[ "${{github.ref_name}}" == 'test-release' ]] ) ; then
+            gh release upload --clobber tip "${artefacts[@]}" || true
```
dskloetd, line 51: Do you really want to update "tip" without first doing a test run on nns-dapp to make sure the snapshot works?

dskloetd: I was thinking we probably don't want to point nns-dapp CI to tip at all but to specific releases. Then we always update and test explicitly. What do you think?

bitdivine: I agree fully - I hadn't submitted my response. :-)

I would like a hefty set of tests for full releases.  Basically, if this claims to be like a fully functional clone of the mainnet NNS, it had better work.  That would mean tests for each canister.  The nns-dapp, the sns canisters and so on.  But that would take too much time for a test run on every commit.

bin/dfx-snapshot-mk-stock
bitdivine: Ok, I have made a script that waits for the next checkpoint and stops the replica only then.  Let's see what that does to start times.

bitdivine: It _looks_ promising, with an 11s start time instead of the 40s start time I have become accustomed to.

.github/workflows/snapshot.yml
bitdivine: Also, not the different filename, so this would not have changed any current CI jobs.  Dependencies should generally be fixed, IMO, with a few exceptions.  Using tip was fine for a proof of concept but we are past that.  The concept has proven its value.  We can put in the engineering hours now to make proper releases and have proper fixed dependencies.

dskloetd (approved):

PR #126, opened 2023-04-20
Flag to disable latest tests

bitdivine:
# Motivation
Form time to time, as now, the latest IC repo is no longer compatible with the snsdemo.  When that happens, the "latest" canary test goes red.  Good, it has done its job.  But once the canary has died there is no point in killing many more canaries until the gas has been cleared from the mine.

# Changes
- Disable the latest test with an "if" so that it is visible in CI that the latest check has been skipped.

# Tests
- See CI. You should see a grayed out "latest" test.  And no failed tests.

### DIFF
--- a/.github/workflows/run.yml
+++ b/.github/workflows/run.yml
@@ -6,7 +6,6 @@ on:
   pull_request:
 env:
   GH_TOKEN: ${{ github.token }}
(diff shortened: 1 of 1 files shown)


dskloetd (approved):

bitdivine:
@lmuntaner Its something done every couple of months or so, depending on priorities.

PR #128, opened 2023-04-21
snsdemo: Add an open swap to the stock snapshot.

dskloetd:
# Motivation

We want to test swap participation in nns-dapp.
For that we need an open swap to participate in.

# Changes

1. Create a sns project that is not finalized.
2. Increase the swap duration to 90 days to make the stock snapshot usable for as long as possible.
3. Add 10 more sns'es because nns-dapp CI currently expects 12 SNS projects.

# Tested

Created and loaded the snapshot locally and checked that it had 11 closed and 1 open swap.

### DIFF
--- a/bin/dfx-sns-sale-propose
+++ b/bin/dfx-sns-sale-propose
@@ -16,8 +16,8 @@ export DFX_NEURON_ID="${DFX_NEURON_ID:-$(dfx-neuron-id --identity "$DFX_IDENTITY
 export DFX_IDENTITY_PEM="$(dfx-identity-pem --identity "$DFX_IDENTITY")"
 export DFX_NNS_URL="$(dfx-network-provider --network "$DFX_NETWORK")"
 
-# The due date of the sale must be greater than 1 day at thetime of NNS proposal execution
-SWAP_DUE_TIMESTAMP="$(perl -e 'print time() + 14 * 24 * 3600')"
+# The due date of the sale must be greater than 1 day at the time of NNS proposal execution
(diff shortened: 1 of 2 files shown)


bin/dfx-sns-sale-propose
```diff
@@ -16,8 +16,8 @@ export DFX_NEURON_ID="${DFX_NEURON_ID:-$(dfx-neuron-id --identity "$DFX_IDENTITY
 export DFX_IDENTITY_PEM="$(dfx-identity-pem --identity "$DFX_IDENTITY")"
 export DFX_NNS_URL="$(dfx-network-provider --network "$DFX_NETWORK")"
 
-# The due date of the sale must be greater than 1 day at thetime of NNS proposal execution
-SWAP_DUE_TIMESTAMP="$(perl -e 'print time() + 14 * 24 * 3600')"
+# The due date of the sale must be greater than 1 day at the time of NNS proposal execution
+SWAP_DUE_TIMESTAMP="$(perl -e 'print time() + 90 * 24 * 3600')"
```
lmuntaner, line 20: Does this mean that we'll need to upgrade the snapshot every 90 days or so?

I guess there isn't a better way, right?

lmuntaner (approved):
Thanks!

bin/dfx-sns-sale-propose
dskloetd: In theory, yes.
For now we'll be upgrading it more often because we want new features.
Hard to tell what the future looks like in 90 days :-).
I tried to keep the swap open even longer but 90 is the maximum that works.
More doesn't cause any errors but the SNS simply doesn't appear.

PR #133, opened 2023-04-21
Set frontend canister IDs and candid

bitdivine:
# Motivation
In the snsdemo we need to add root  as a controller of the frontend dapps.  There is a problem that needs to be fixed first though: The stock nns frontend dapps are installed by `dfx nns install` (and later updated to the version we actually  want to test).  The canister IDs are meant to be set by `dfx nns import` but the latter doesn't set the canister IDs for the frontend dapps.  It is hard to transfer the ownership of a canister, the ID of which is not  known.  Yes, we can hard-wire a canister ID but this is fragile.

I have a made a PR to fix this upstream.  However it typically takes weeks or more for such a change to trickle down to us.  In the meantime, this PR extends the `dfx-nns-import` polyfill with the missing functionality.

# Changes
- Set the canister ID and candid for the nns-dapp in `dfx-nns-import`.
- Remove places where the canister IDs were already hard-wired.

# Tests
I have added a test for the `dfx-nns-import` polyfill.  It should also be possible to verify that the upstream fix works as we expect by applying the same test to `dfx nns import`.

### DIFF
--- a/bin/dfx-ckbtc-deploy
+++ b/bin/dfx-ckbtc-deploy
@@ -103,7 +103,7 @@ deploy_ckbtc() {
 check_ckbtc_deployment() {
   for canister in "${LOCAL_PREFIX}ledger" "${LOCAL_PREFIX}minter"; do
     : "Verify that the deployed canister matches the local wasm"
-    dfx-canister-check-wasm-hash --canister "$canister"
+    dfx-canister-check-wasm-hash --canister "$canister" --network "$DFX_NETWORK"
   done
 }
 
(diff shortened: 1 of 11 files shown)


bin/dfx-ckbtc-deploy
bitdivine, line 1: From stacked PR.

bin/dfx-stock-deploy
bitdivine, line 1: From stacked PR.

bin/dfx-stock-deploy
```diff
@@ -0,0 +1,47 @@
...
+: Add 10 more finalized SNS projects.
+dfx-sns-demo-mksns-parallel --network "$DFX_NETWORK" --num_sns 10 --majority snsdemo8
+
+: Add 1 open SNS project.
+dfx-sns-demo-mksns --network "$DFX_NETWORK"
+
+: Set up ckbtc canisters
+dfx-ckbtc-import --prefix ckbtc_
+dfx-ckbtc-deploy --prefix ckbtc_ --network "$DFX_NETWORK" --yes
+
```
bitdivine, line 42: New code: Add controller

mstrasinskis (approved):
Rubber stamp.

PR #125, opened 2023-04-20
Mv stock state creation to a separate script

bitdivine:
# Motivation
It would be nice to be able to deploy testnets with the stock state.

# Changes
- Move the stock state creation into a separate script
- Fix a missing `--network` flag.

# Tests
- Local deployments - See CI
- Testnet deployments: Deployed to small09

PR #134, opened 2023-04-24
Let root control frontend dapps

bitdivine:
# Motivation
Currently, in the demo, the `nns-dapp` and `internet_identity` are not controlled by the root, so they cannot be upgraded by proposal.

Note: Stacked on #125

# Changes
Add the root as a controller of these canisters


mstrasinskis (approved):
LGTM

PR #136, opened 2023-04-25
Create the open swap among the first 10.

dskloetd:
# Motivation

nns-dapp currently reads only the first page from the sns_aggregator.
It used to be the last page but we need a project which has tokens in the faucet and currently only the first project (created by `bin/dfx-sns-demo`) has tokens in the faucet so we changed it to reading the first page.
I'm adding an e2e test which requires an open swap.
So for nns-dapp to be able to see the open swap (while using the sns_aggregator), it needs to be among the first 10 projects.

# Changes

Create the open swap before creating the 10 additional projects.

# Testing

1. Install sns_aggregator locally to reproduce the problem.
2. Created a new stock snapshot with this change.
3. Used the stock snapshot with the sns_aggregator running to see that an open swap is now available.


mstrasinskis (approved):
Rubber stamp. Thanks!

lmuntaner (approved):

PR #138, opened 2023-04-26
Delete unused dfx nns install

bitdivine:
# Motivation
There are scripts that were used by the last demo that are no longer needed.

# Changes
- Delete some scripts that are unused.

# Tests
See CI


bin/dfx-canister-import
dskloetd, line 1: You said you might want to keep this one.

dskloetd (approved):

PR #140, opened 2023-04-27
Test snapshots from a clean environment

bitdivine:
# Motivation
Some canister IDs are not exported for snapshots.  This is _probably_ because canister IDs may be set in three different places and it is set in the "wrong" place.  To test this properly, it is necessary to test from a clean state


bitdivine:
Outdated

PR #142, opened 2023-04-27
When installing dfx, make sure the cache is populated.

bitdivine:
# Motivation
The dfx cache is not installed automatically when dfx is installed.  It should be.

# Changes
- In `dfx-software-dfx-install`, install the cache as well.

# Tests
See CI

### DIFF
--- a/bin/dfx-software-dfx-install
+++ b/bin/dfx-software-dfx-install
@@ -14,8 +14,14 @@ source "$(clap.build)"
 [[ "${DFX_VERSION}" != "latest" ]] || DFX_VERSION=""
 DFX_VERSION="${DFX_VERSION#v}"
 
+# Install the dfx cache if it hasn't been installed already
+lazy_dfx_cache_install() {
+  test -d "$(dfx cache show)" || dfx cache install
+}
+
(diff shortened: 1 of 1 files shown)


bin/dfx-software-dfx-install
```diff
@@ -14,8 +14,14 @@ source "$(clap.build)"
 [[ "${DFX_VERSION}" != "latest" ]] || DFX_VERSION=""
 DFX_VERSION="${DFX_VERSION#v}"
 
+# Install the dfx cache if it hasn't been installed already
+lazy_dfx_cache_install() {
```
dskloetd, line 18: What do you mean by lazy?
In my mind it would be lazy if you only install it when you need it (like you do in `bin/dfx-sns-wasm-upload`).
Here it looks proactive to me.
But maybe you're referring to something else?

bitdivine: I mean that it doesn't do unnecessary work installing the cache if it is already installed.  For sure it doesn't correspond to the functional programmer's definition of lazy but it matches the Oxford definition. :-)

bitdivine: For contrast, `dfx cache install` will reinstall the cache, even if it is already there.

dskloetd (approved):

PR #144, opened 2023-04-28
Add tools needed to work with the farm

bitdivine:
# Motivation
I am getting deployments to the farm working cleanly.  I am encoding the steps needed to get there.  These steps are almost certain to change as I learn more.

# Changes
- Add a script to install the tools I needed.  Sorry, Linux only.  Not sure whether this is even possible on Mac.  Linux dev servers are available.

# Tests
None


dskloetd:
What is a farm?

bitdivine:
@dskloetd Farm is the name of a shiny new system for spinning up testnets on the fly.  Currently aimed at short term use, about an hour, but it might be expanded to cover our use cases.

PR #146, opened 2023-04-28
Install wasm nm

bitdivine:
# Motivation
In https://dfinity.atlassian.net/browse/GIX-1521, @dskloetd found that wasm-nm is expected but not available.

# Changes
- Add an install script for wasm-nm
- Add a wrapper for canister wasms, as we always use the same incantation.  At least, I do.

# TODO
Find all places where this is relevant and insert it.

# Tests
Rollback and forwards:
```
$ chronic ./bin/dfx-software-wasm-nm-install --version 0.1.0 && wasm-nm --version && chronic ./bin/dfx-software-wasm-nm-install --version 0.2.0 && wasm-nm --version
wasm-nm 0.1.0
wasm-nm 0.2.0
```

Check that the wrapper parses canisters:
```
bin/dfx-canister-nm ../../../nns-dapp/nns-dapp.wasm 
canister_query http_request
canister_query get_account
canister_update add_account
canister_query get_transactions
[SNIP]
```
Note:  Running on all wasms found on my  machine found some that failed to be parsed by wasm-nm.  These use simd instructions that are not supported by wasm-nm:
```
max@sinkpad:~/dfn/snsdemo/branches/install-wasm-nm (1:45:09)$ locate -r '\.wasm$' | head | xargs -I{} ./bin/dfx-canister-nm  {}
[SNIP]
error: /home/max/.cache/bazel/_bazel_max/11b917098ce4ca861488a79c139364ca/execroot/ic/bazel-out/k8-opt/bin/external/sw_npm/node_modules/@dfinity/response-verification/debug/nodejs/nodejs_bg.wasm: Unknown opcode 192
[SNIP]
```

### DIFF
--- a/bin/dfx-canister-nm
+++ b/bin/dfx-canister-nm
@@ -1,0 +1,27 @@
+#!/usr/bin/env bash
+set -euo pipefail
+SOURCE_DIR="$(dirname "$(realpath "${BASH_SOURCE[0]}")")"
+PATH="$SOURCE_DIR:$PATH"
+. "$SOURCE_DIR/versions.bash"
+
+print_help() {
+  cat <<-EOF
+
+	Print functions exported by a canister.
+	EOF
+}
+
+# Source the clap.bash file ---------------------------------------------------
+source "$SOURCE_DIR/clap.bash"
+# Define options
+# Source the output file ----------------------------------------------------------
+source "$(clap.build)"
+
+# The only meaningful export from canister wasms is function names.
+canister_nm=(wasm-nm -je)
+
+canister_wasm="$1"
+case "$(file "$canister_wasm")" in
+*gzip*) "${canister_nm[@]}" <(gunzip <"$canister_wasm") ;;
+*) "${canister_nm[@]}" "$canister_wasm" ;;
+esac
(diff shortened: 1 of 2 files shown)


bin/dfx-canister-nm
```diff
@@ -0,0 +1,27 @@
...
+# Source the clap.bash file ---------------------------------------------------
+source "$SOURCE_DIR/clap.bash"
+# Define options
+# Source the output file ----------------------------------------------------------
+source "$(clap.build)"
+
+# The only meaningful export from canister wasms is function names.
+canister_nm=(wasm-nm -je)
+
+canister_wasm="$1"
```
dskloetd, line 23: Does clap provide help on positional args or should it be added manually?

bitdivine: It doesn't.  I'd like to add support for positional args, as it would give us automatic documentation.  At one time I was hesitant to add support as array handling, limited as it is, is not supported in decade old macbash.  But if we have a policy of getting people to install an up to date bash, that reason disappears.

dskloetd: In that case I think it's worth adding a usage line to `print_help` so it's clear how the parameter is passed.

dskloetd (approved):

dskloetd:
Can you respond to my comment?

PR #148, opened 2023-05-02
snsdemo: Script to install internet identity

dskloetd:
# Motivation

Internet identity stopped providing downloadable artifacts which are not gzipped, but in the past didn't provide downloadable artifacts which are gzipped.

Note that `dfx-software-internet-identity-version` doesn't actually work the way it's currently used by `bin/dfx-software-internet-identity-ci-wasm-url` but that's a separate issue which I'll discuss with Max when he's back. That's not blocking us right now.

# Changes

1. Change `bin/dfx-software-internet-identity-ci-wasm-url` to add `.gz` to the URL if the release is after `release-2023-04-12`
2. Add `bin/dfx-software-internet-identity-install` which installs internet identity and which understands that the downloaded artifact

# Tested

Manually hard coded different release names and checked that ii was installed with and without `.gz`.


frederikrothenberger (approved):
LGTM

PR #151, opened 2023-05-08
Remove duplicate installation of internet_identity

dskloetd:
# Motivation

`bin/dfx-network-deploy` has 2 blocks of code to install `internet_identity`. One is conditional and one always installs it.

# Changes

Remove the code that conditionally installs internet_identity since it will be installed again anyway.

# Tests

CI.


bitdivine (approved):
LGTM.  Thanks.

PR #155, opened 2023-05-09
Clean up dfx nns import commands

bitdivine:
# Motivation
There is a polyfill for dfx nns import, however it is applied inconsistently.

# Changes
- Pass polyfill arguments through, so that the polyfill can be used for networks other than local.
- Apply the polyfill in more places.

# Tests
- See CI for local deployments
- The small09 testnet was deployed with these changes.


bitdivine:
Outdated

PR #156, opened 2023-05-09
Deploy aggregator

bitdivine:
# Motivation
We would like to deploy the aggregator along with the nns-dapp

# Changes
- Deploy the aggregator

# TODO
- Deploy the frontend dapps using prebuilt binaries
- Exit test for the stock state, including verifying that this canister is present, the canister ID is defined and there is an entry in dfx.json.

### DIFF
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -96,6 +96,5 @@ jobs:
         run: |
           set -x
           mkdir ~/dfx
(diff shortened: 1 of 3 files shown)


dskloetd (approved):

bitdivine:
Note: I have added tests and a fix without which this change is useless.  This doe doesn't apply to the local testnet.

I cannot push right now - is github down?  Once it is up I will ask for a re-review.

PR #158, opened 2023-05-09
Download lifeline.wasm in dfx-nns-install

dskloetd:
# Motivation

We want to be able to bump the IC commit and run CI against the latest IC commit.
Currently `dfx nns install` fails with newer commits because it tries to download `lifeline.wasm` from a URL that doesn't exist for newer IC commits.

# Changes

1. In `bin/dfx-nns-install` try to prepopulate the dfx wasms cache with `lifeline.wasm`
2. Enable running CI at the latest IC commit once again.

# Tests

See CI


bitdivine:
Interestingly this too doesn't appear with this filter:
is:pr is:open review:required draft:false status:success 

It seems to tick every box and yet it doesn't appear.

dskloetd:
Maybe they didn't appear because I pushed them while GitHub was broken and it failed to update some index?

bitdivine (approved):
LGTM.  Thanks.  Yay, we have latest working again.

dskloetd:
If I remove `review:required`, it appears in the search for me:
https://github.com/dfinity/snsdemo/pulls?q=is%3Apr+is%3Aopen+draft%3Afalse+status%3Asuccess+
Which makes sense because they are approved now, I guess?

PR #159, opened 2023-05-09
snsdemo: Bump IC commit

dskloetd:
# Motivation

I'm having issues with the SNS participation end-to-end test on nns-dapp CI and I'm wondering if the canister versions have something to do with it.
Later we'll also have to update the commit again to test changes for geo-restriction, and it's good to already be less behind when we do.

# Changes

Bump the default IC commit to 89129b8212791d7e05cab62ff08eece2888a86e0. This is the same commit we use for dependencies in the `Cargo.toml` of the nns-dapp. It is from 2023-04-25.

# Tests

See CI

### DIFF
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -76,3 +76,13 @@ jobs:
           git clean -dfx
           bin/dfx-ckbtc-deploy.test
           git clean -dfx
+  nns-dapp-canister-checks:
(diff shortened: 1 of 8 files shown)


bitdivine:
Interestingly this PR did not appear with this filter:
is:pr is:open review:required draft:false status:success 

I only saw it because I looked at the full list of PRs.

PR #160, opened 2023-05-10
feat: Command to get nns-dapp versions

bitdivine:
# Motivation
We wish to install the nns-dapp wasms from prebuilt releases.  For this, we need to know which version of nns-dapp to install

Note: This is similar to the code for other canisters such as `internet_identity` version command and the (get to be merged) `sns_aggregator` version command, however on close examination each has details that differ so combining them is probably not worthwhile at present.

# Changes
- Define a default (pinned) version 
- Define a command to get the pinned release or the latest release by proposal.

# Tests
A test has been added and is exercised in CI

### DIFF
--- a/bin/dfx-software-nns-dapp-version.test
+++ b/bin/dfx-software-nns-dapp-version.test
@@ -1,0 +1,46 @@
+#!/usr/bin/env bash
+set -euo pipefail
+SOURCE_DIR="$(dirname "$(realpath "${BASH_SOURCE[0]}")")"
+PATH="$SOURCE_DIR:$PATH"
+
+(
+  echo "Using --pinned should get a release from the versions file"
+  ACTUAL_RELEASE="$(dfx-software-nns-dapp-version --pinned)"
+  EXPECTED_RELEASE="$(. "$SOURCE_DIR/versions.bash" && echo "${NNS_DAPP_RELEASE}")"
+  [[ "$ACTUAL_RELEASE" =~ ^proposal-[0-9]*$ ]] || {
+    echo "ERROR: The release is not a proposal tag: '$ACTUAL_RELEASE'"
+    exit 1
+  } >&2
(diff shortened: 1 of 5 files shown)


bin/dfx-software-nns-dapp-version.test
```diff
@@ -0,0 +1,46 @@
...
+set -euo pipefail
+SOURCE_DIR="$(dirname "$(realpath "${BASH_SOURCE[0]}")")"
+PATH="$SOURCE_DIR:$PATH"
+
+(
+  echo "Using --pinned should get a release from the versions file"
+  ACTUAL_RELEASE="$(dfx-software-nns-dapp-version --pinned)"
+  EXPECTED_RELEASE="$(. "$SOURCE_DIR/versions.bash" && echo "${NNS_DAPP_RELEASE}")"
+  [[ "$ACTUAL_RELEASE" =~ ^proposal-[0-9]*$ ]] || {
+    echo "ERROR: The release is not a proposal tag: '$ACTUAL_RELEASE'"
```
dskloetd, line 11: I think it's good practice for such messages to say "... should ...".
Otherwise it can be unclear whether the message describes the correct or the incorrect state.
So:
ERROR: The release should be a proposal tag: '$ACTUAL_RELEASE'

bin/dfx-software-nns-dapp-version.test
```diff
@@ -0,0 +1,46 @@
...
+    exit 1
+  } >&2
+  [[ "$ACTUAL_RELEASE" == "$EXPECTED_RELEASE" ]] || {
+    echo "ERROR: The release does not correspond to a naive reading of the versions file."
```
dskloetd, line 15: ERROR: The release should correspond to a naive reading of the versions file.

dskloetd (approved):

PR #163, opened 2023-05-10
Clear dfx wasms cache before 'dfx nns install'

dskloetd:
# Motivation

When running `dfx nns install` you can specify a `DFX_IC_COMMIT` to determine which versions of canisters will be installed. But if canisters are already present in `$(dfx cache show)/wasms`, those will be used regardless of which version they are.
This results in unpredictable canister versions being installed on a manually created snapshot.
We would like a manually created snapshot to mirror CI as close as possible to we can test locally under conditions as similar as possible to CI.

# Changes

1. In `bin/dfx-nns-install`, empty the dfx wasms cache directory before running `dfx nns install`.
2. Move the removed wasms back to the cache afterwards.

I intentionally didn't remove wasms that weren't there before. This is debatable but I think it might be useful to have them and it's most important that the snapshot itself is predictable. Let me know if you disagree.

# Tests

Ran `bin/dfx-nns-install` (both with empty and non-empty wasms cache) and watched it empty the cache, download wasms and restore them afterwards.

### DIFF
--- a/bin/dfx-nns-install
+++ b/bin/dfx-nns-install
@@ -26,8 +26,9 @@ clap.define short=c long=ic_commit desc="The IC commit of the wasms" variable=DF
 source "$(clap.build)"
 
 WASMS_DIR="$(dfx cache show)/wasms"
+mkdir -p $WASMS_DIR
 
-if [ -d $WASMS_DIR ]; then
+if [ "$(ls -A "$WASMS_DIR")" ]; then
   # 'dfx nns install' will use wasms present in $WASMS_DIR, even if they do not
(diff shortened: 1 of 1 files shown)


bitdivine (approved):
Looks good, thanks.  Personally, I would just delete the wasms dir.  It is a cache, after all.  Emptying a cache shouldn't break anything and `rm -fr "$WASMS_DIR" ; mkdir "$WASMS_DIR"` is simpler.  But is this works I'm fine with it.

dskloetd:
I might agree in the future but for now I prefer to keep it.

PR #164, opened 2023-05-10
Aggregator download url

bitdivine:


bin/dfx-software-internet-identity-ci-wasm-url
bitdivine, line 1: Tweak: The --verbose creates a lot of spam.

bin/dfx-software-sns-aggregator-ci-wasm-url
```diff
@@ -0,0 +1,42 @@
...
+	Outputs the URL of the wasm file for the sns_aggregator canister.
+
+	EOF
+}
+
+# Source the clap.bash file ---------------------------------------------------
+source "$SOURCE_DIR/clap.bash"
+# Define options
+clap.define short=r long=release desc="The release name" variable=DFX_RELEASE default="pinned"
+# TODO: Add documentation similar to: https://github.com/dfinity/internet-identity#flavors
```
bitdivine, line 19: Note: This TODO is being addressed with the PR against the nns-dapp repo.

bin/dfx-software-sns-aggregator-ci-wasm-url
```diff
@@ -0,0 +1,42 @@
...
+# TODO: Write a virus that replaces American english with British english worldwide.
```
dskloetd, line 20: Did you mean "wourldwide" instead of "worldwide"?

bin/dfx-software-sns-aggregator-ci-wasm-url
```diff
@@ -0,0 +1,42 @@
...
+
+	Outputs the URL of the wasm file for the sns_aggregator canister.
+
+	EOF
+}
+
+# Source the clap.bash file ---------------------------------------------------
+source "$SOURCE_DIR/clap.bash"
+# Define options
+clap.define short=r long=release desc="The release name" variable=DFX_RELEASE default="pinned"
```
dskloetd, line 18: It's the aggregator release, not the DFX release, right? Or what is the `DFX_` prefix signifying here?

bin/dfx-software-sns-aggregator-ci-wasm-url.test
```diff
@@ -0,0 +1,45 @@
...
+  PROD_URL="$(dfx-software-sns-aggregator-ci-wasm-url --flavor prod)"
+  [[ "${DEFAULT_URL:-}" == "${PROD_URL}" ]] || {
+    echo "ERROR: The default URL should be the production URL."
+    echo "  DEFAULT:    '${DEFAULT_URL}'"
+    echo "  PRODUCTION: '${PROD_URL}'"
+    exit 1
+  } >&2
+)
+
+echo "$(basename "$0") PASSED"
```
dskloetd, line 45: If I'm not mistaken, if the `--flavor` flag does literally nothing, everything in this test would still pass.
Maybe test that the `dev` URL actually has `_dev` in the wasm file name while the `prod` URL doesn't?

In general I think it's better to assert explicit values in tests rather than asserting that one generated thing is equal or different from another generated thing because it makes it easier to see what is actually expected and less susceptible to both generators failing in the same way.

bin/dfx-software-sns-aggregator-ci-wasm-url
bitdivine: My hope is that many of these scripts are polyfills and will eventually make their way into dfx core or dfx extensions.  The DFX_ prefix to inputs follows from that hope.    dfx has traditionally takes some inputs as flags, others as environment variables.  The dfx executable names its environment variables DFX_xxx.  If a polyfill is implemened in dfx, we can first call it directly in the polyfill, fixing up and differences as needed, then eventually eliminate the polyfill.  Hopefully the variable names will be compatible enough that the implementer will not be tempted to change them.

bitdivine: That said, the name is confusing, I agree.  Something like `DFX_AGGREGATOR_RELEASE` would be better.

bitdivine: Done.

bin/dfx-software-sns-aggregator-ci-wasm-url
bitdivine: just  frustration with things like flavors and artifact.  I generally like having many different dialects and local flavours.  I just don't like being forced to use one that isn't my own.

bin/dfx-software-sns-aggregator-ci-wasm-url.test
```diff
@@ -0,0 +1,69 @@
...
+    dfx-canister-nm "$DOWNLOAD" | grep -q canister_query || {
+      echo "ERROR: The downloaded file should be a canister wasm file but doesn't appear to be."
+      echo "       It should be:"
+      echo "       - A wasm file"
+      echo "       - With canister query calls"
+      echo "       - Optionally gzipped"
+      echo "Download file type:"
+      file "$DOWNLOAD"
+      echo "File stats:"
+      ls -l "$DOWNLOAD"
```
bitdivine, line 51: Note: The URL check fails sometimes.  Why, I have no idea.  The URL seems valid, it looks as if curl "succeeds" but the file is invalid.  Of course when one adds diagnostics the problem disappears, but whenever it reappears it would be good to have a lot of details on what is happening.

dskloetd: It might be that CI gets a boundary node in the US while you get one in Europe. Maybe try while connecting your VPN to the US?

At least that we had issues like that while you were away where 1 US boundary node was broken and CI failed 33% of the time.

dskloetd (approved):

PR #165, opened 2023-05-11
Deploy aggregator standalone

bitdivine:
# Motivation
We would like to deploy the aggregator to the stock state.  Doing this by pulling in the nns-dapp repo and moving configuration around turned out to be extremely flaky.  Instead, this provides a tool for installing the sns_aggregator from a pre-built binary.

# Changes
- Add a command to install the sns_aggregator canister and corresponding dfx.json config.

# Tests
- A test is included, and it is exercised in CI.

### DIFF
--- a/bin/dfx-software-sns-aggregator-install.test
+++ b/bin/dfx-software-sns-aggregator-install.test
@@ -16,29 +16,33 @@ function cleanup {
 }
 trap cleanup EXIT
 
-(
-  echo "Should install the sns_aggregator"
-  : Clean up
-  git checkout dfx.json
-  dfx-network-stop
-  dfx start --clean --background 2>/dev/null
-  : Install
-  dfx-software-sns-aggregator-install
-  : The downloaded wasm should be the pinned, production version, not called 'pinned' or some mutable alias:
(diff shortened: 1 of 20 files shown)


bin/dfx-software-sns-aggregator-install.test
```diff
@@ -0,0 +1,48 @@
...
+      : Clean up
+      git checkout dfx.json
+      dfx-network-stop
+      dfx start --clean --background 2>/dev/null
+      : Install
+      dfx-software-sns-aggregator-install --release "$release" --flavor "$flavour"
+      : The downloaded wasm should be the pinned, production version, not called 'pinned' or some mutable alias:
+      SNS_AGGREGATOR_RELEASE="$(. "$SOURCE_DIR/versions.bash" && echo "$SNS_AGGREGATOR_RELEASE")"
+      [[ "$SNS_AGGREGATOR_RELEASE" =~ ^proposal-[0-9]+-agg$ ]] || {
+        echo "ERROR: The pinned $release $flavour release should refer to a proposal: '$SNS_AGGREGATOR_RELEASE'"
```
dskloetd, line 32: Why does it say `pinned` here? This code path is taken for each of the values or $release, right?

bitdivine: Good catch.

dskloetd (approved):

PR #168, opened 2023-05-13
Fix: Newline after wait

bitdivine:
# Motivation
Fix a niggling error with dfx-network-stop:
- It prints an unterminated line, so the next output is mixed in with this output
```
Stopping icx-proxy...
Stopped.
   99Running dfx start for version 0.13.1
     ^ Start of next output
```
This is caused by the wait loop that repeatedly writes the same line; it should be followed by something that ends in a new line.
# Changes
- Print after the wait loop.

### DIFF
--- a/bin/dfx-network-stop
+++ b/bin/dfx-network-stop
@@ -16,10 +16,11 @@ if [[ "$DFX_NETWORK" == "local" ]]; then
   pgrep --full '\<dfx\>.*\<start\>' | xargs --no-run-if-empty kill || true
   echo "Waiting for replica to stop..."
   for ((i = 100; i > 0; i--)); do
-    printf '\r % 4d' "$i"
     pgrep replica || break
+    printf '\r % 4d' "$i"
     sleep 1
   done
+  echo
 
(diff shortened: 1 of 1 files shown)


bin/dfx-network-stop
```diff
@@ -17,9 +17,10 @@ if [[ "$DFX_NETWORK" == "local" ]]; then
   echo "Waiting for replica to stop..."
   for ((i = 100; i > 0; i--)); do
     printf '\r % 4d' "$i"
-    pgrep replica || break
+    pgrep replica 2>/dev/null || break
```
bitdivine, line 20: Printing the replica pid is incompatible with the progress printout that is trying to reuse the line.

bin/dfx-network-stop
```diff
@@ -17,9 +17,10 @@ if [[ "$DFX_NETWORK" == "local" ]]; then
...
     sleep 1
   done
+  printf "\n...stopped.\n"
```
bitdivine, line 23: This ends the reused line.

dskloetd (approved):

PR #172, opened 2023-05-19
snsdemo: Add confirmation_text to open SNS swap

dskloetd:
# Motivation

SNS swaps support a new field `confirmation_text` which is something participants need to agree to.
We want to be able to test the field in nns-dapp.

`quill` does not yet support passing `--confirmation_text` so we can use the field only for projects that we leave open and don't participate in from snsdemo.

Until we update our pinned IC_COMMIT, the new field is only taken into account when using `--ic_commit latest`.

# Changes

1. Add commented out `confirmation_text` field to `bin/sns_init.yml`.
2. Support `--confirmation_text` in `bin/dfx-sns-config-random` to use the `confirmation_text` field when generating `sns.yml` from `bin/sns_init.yml`.
3. Stop passing `--network` to `bin/dfx-sns-config-random`. The script does not support this flag and once it uses `clap` it will fail on unsupported flags.
4. Support `--confirmation_text` in `bin/dfx-sns-demo-mksns` by passing it on to `bin/dfx-sns-config-random`.
5. In `bin/dfx-snapshot-stock-make` pass `--confirmation_text` to `bin/dfx-sns-demo-mksns`, only for the SNS swap that will remain open.
6. Support `--ic_commit` and `--ic_dir` in `bin/dfx-snapshot-stock-make` to be able to create a snapshot with the latest canisters.

# Tested

I created a new stock snapshot with
```
./bin/dfx-snapshot-stock-make -s ~/snapshots/confirmation_text-2023-05-19.tar.xz --ic_commit latest -x ../../ic
```
and ran it with sns_aggregator and nns-dapp to verify that it includes the confirmation text.

I also created a snapshot without `--ic_commit latest` and verified that it works but does not include the confirmation text.

### DIFF
--- a/bin/dfx-sns-config-random
+++ b/bin/dfx-sns-config-random
@@ -1,5 +1,20 @@
 #!/usr/bin/env bash
 set -euo pipefail
+SOURCE_DIR="$(dirname "$(realpath "${BASH_SOURCE[0]}")")"
+
+print_help() {
+  cat <<-EOF
+
+	Creates an sns.yml to configure an SNS swap.
+	EOF
+}
+
+# Source the clap.bash file ---------------------------------------------------
+source "$SOURCE_DIR/clap.bash"
+# Define options
+clap.define long=confirmation_text desc="A text that needs to be confirmed by the swap participants" variable=CONFIRMATION_TEXT default=""
+# Source the output file ----------------------------------------------------------
+source "$(clap.build)"
 
 consonant() {
   openssl rand -base64 100 | tr -dc BCDFGHJKLMNPQRSTVWXZ | head -c 1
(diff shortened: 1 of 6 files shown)


bin/dfx-sns-config-random
```diff
@@ -21,6 +36,15 @@ PRINCIPAL="$(dfx identity get-principal)"
...
 # dfx canister info smiley_dapp --network "$DFX_NETWORK" | perl -e 'while($_ = <>){ if (s/^Controllers: *//){s/ +/, /g;print $_} }'
 
-sed "s/TOKEN_NAME/$TOKEN_NAME/g;s/TOKEN_SYMBOL/$TOKEN_SYMBOL/g;s/PRINCIPAL/$PRINCIPAL/g" "$(dirname "$(realpath "$0")")"/sns_init.yml >sns.yml
+SNS_INIT_PATH="$(dirname "$(realpath "$0")")/sns_init.yml"
+SED_COMMANDS="s/TOKEN_NAME/$TOKEN_NAME/g;
+              s/TOKEN_SYMBOL/$TOKEN_SYMBOL/g;
+              s/PRINCIPAL/$PRINCIPAL/g"
+
+if [ "${CONFIRMATION_TEXT:-}" ]; then
+  SED_COMMANDS="$SED_COMMANDS;s/# confirmation_text: \"CONFIRMATION_TEXT\"/confirmation_text: \"${CONFIRMATION_TEXT}\"/"
```
lmuntaner, line 45: Does this add the "confirmation_text" field in sns_yaml.init and "uncomments it?

lmuntaner (approved):
One question, but LGTM. We can merge and ask @bitdivine to review next week.

bin/dfx-sns-config-random
dskloetd: Yes, but it does it by taking all of
```
# confirmation_text: "CONFIRMATION_TEXT"
```
and replacing it with
```
confirmation_text: "${CONFIRMATION_TEXT}"
```
It doesn't separately remove `#` and fill in the placeholder or anything like that.

In general `sed 's/foo/bar/'` replaces foo with bar.

I'll add a comment.

lmuntaner: Thanks for the explanation!

bitdivine (approved):

bin/dfx-sns-config-random
dskloetd: Thanks, I agree it's fine to keep it simple as it is.

PR #173, opened 2023-05-23
Make dfx-software-dfx-install retry longer and fail properly

dskloetd:
# Motivation

The `dfx` install script is served from the IC.
The canister that serves it is upgraded "~a dozen times a day", and this is (probably) causing the 500 errors we are seeing in CI.
Also when this happens, the error doesn't cause CI to fail but it fails only later when something else tries to run `dfx`

# Changes

1. Do more retries (up to 8 minutes) to download the install script.
2. Download the install script to a file and execute the file, in order to properly notice when downloading fails.

# Tests

None


bitdivine (approved):
LGTM, thanks!

PR #174, opened 2023-05-23
Fix dfx-network-stop for Mac

dskloetd:
# Motivation

On Mac:
* The flag `--full` doesn't exist for `pgrep`
* The flag `--no-run-if-empty` doesn't exist for xargs
* The `\<` word boundary regex doesn't work

# Changes

1. Add back `dfx stop` to `bin/dfx-network-stop`
2. Use `pgrep -f` instead of `pgrep --full` and `xargs -n 1` instead of `xargs --no-run-if-empty` such that even though it still doesn't help on Mac (because of the regex), it would cause errors.

# Tests

Tested manually on Mac


bitdivine (approved):
This really needs to be fixed upstream.  The energy we're consuming downstream isn't cool.

PR #175, opened 2023-05-24
GIX-1561 snsdemo: Bump IC commit

dskloetd:
# Motiviation

We need to create an SNS with `confirmation_text` and `restricted_countries`, which were added after the currently pinned IC commit.

# Changes

1. Change the IC commit to one after https://github.com/dfinity/ic/commit/9b0e2417043ea16e5bd7c9e12e7178a85e0d76fa which added the possibility to add these fields to the `sns.yml` file.
2. Change  `rs/rosetta-api/icrc1/ledger/icrc1.did` to `rs/rosetta-api/icrc1/ledger/ledger.did` in `bin/dfx-ckbtc-import` because the former no longer exists at the new commit.
3. Do not skip downloading `.did` files if they are already present, in order to guarantee that we get the version from the commit we requested.

# Tests

Created a snapshot manually on M1 Mac and on GitHub CI.

### DIFF
--- a/bin/dfx-ckbtc-import
+++ b/bin/dfx-ckbtc-import
@@ -67,16 +67,12 @@ ckbtc_import() {
     local remote_name="$1"
     local local_name="$2"
     local local_path="$(c="$local_name" jq -re '.canisters[env.c].candid' dfx.json)"
-    if test -e "$local_path"; then
-      echo "Skipping $local_path as it already exists"
-    else
-      local url="https://raw.githubusercontent.com/dfinity/ic/$DFX_IC_COMMIT/${remote_name}"
-      echo "Getting  $local_path from $url..."
-      curl -sSLf --retry 5 "$url" -o "$local_path"
-    fi
+    local url="https://raw.githubusercontent.com/dfinity/ic/$DFX_IC_COMMIT/${remote_name}"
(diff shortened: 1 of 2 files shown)


bin/dfx-ckbtc-import
```diff
@@ -67,16 +67,12 @@ ckbtc_import() {
     local remote_name="$1"
     local local_name="$2"
     local local_path="$(c="$local_name" jq -re '.canisters[env.c].candid' dfx.json)"
```
bitdivine, line 69: This has a performance impact.  Why not download to a location that has the version in the path, then copy to wherever you need it?  That too would guarantee that you have the desired version, even if the download is skipped.

bitdivine (approved):
I think there is no need to skip caching; if you are concerned about getting the correct version, that can be solved in other ways.  However that is not a blocker.

bin/dfx-ckbtc-import
dskloetd: I guess my use case might be different than yours? I just want to create a snapshot occasionally, which is slow anyway and then only use the snapshot. So I have no need for the caching.
While I agree your suggestion would allow both caching and correct versions, it also increases complexity and effort.
If you think it's worth it, do you mind doing it?

PR #176, opened 2023-05-24
snsdemo: Document snapshot customization

dskloetd:
# Motivation

It can be useful to make snapshots that diverge from the stock snapshot for specific tasks.

# Changes

Expand the section in the `README.md` about manual snapshot creation.


lmuntaner (approved):
Thanks!

PR #178, opened 2023-05-25
snsdemo: Put the open swap last

dskloetd:
# Motivation

nns-dapp loads only the first page from the sns_aggregator.
It used to load only the last page and we want to change it back to loading the last page.
Eventually we should use pagination but we are not there yet.

# Changes

Create the open swap after creating the other 11 SNSes.


lmuntaner (commented):
I would wait until we decide to add pagination or not.

dskloetd:
Why? Using the first page was already a work-around. We are just switching to a different work-around, which is more convenient during testing but not worse in any way

lmuntaner (approved):
The PR using pagination is in.

This should not break the CI now.

Approving.

dskloetd:
Thanks, but it's also no longer needed because of that.

PR #182, opened 2023-06-01
Remove shellcheck exception

bitdivine:
# Motivation
By default shellcheck encourages separating variable assignment from scoping rules such as `export` or `local`.  To date this rule has been ignored in this repo but maybe shouldn't be.

# Changes
- Enable shellcheck `SC2155`
- Update the code to match

### DIFF
--- a/bin/dfx-canister-url
+++ b/bin/dfx-canister-url
@@ -108,7 +108,8 @@ inject_canister_id() {
   test -n "${root_url:-}" || return
   case "$DFX_URL_TYPE" in
   static)
-    local CANISTER_ID="$(canister_id)"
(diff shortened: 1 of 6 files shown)


bin/dfx-canister-url
dskloetd, line 1: Is this just a matter of taste or is there a reason that this is better/safer/whatever?

dskloetd (approved):

bin/dfx-canister-url
bitdivine:
```
$ x="$(false)" ; echo $?
1
$ export x="$(false)" ; echo $?
0
```

PR #183, opened 2023-06-02
Fewer lint exceptions

bitdivine:
# Motivation
daily cleanup:  When importing snsdemo commands to nns-dapp I sometimes hit the linter in CI.  These are due to insignificant issues, but this still causes friction.  So today's 15 minute tidy goes to reducing the number of shellcheck exceptions in this repo.

# Changes
- Remove 3 linter rules and make the corresponding changes to the code.
- Remove one unused line, that gets unneeded version info.

# Tests
- Existing tests should suffice

### DIFF
--- a/bin/dfx-canister-url
+++ b/bin/dfx-canister-url
@@ -2,7 +2,6 @@
 set -euo pipefail
 SOURCE_DIR="$(dirname "$(realpath "${BASH_SOURCE[0]}")")"
 PATH="$SOURCE_DIR:$PATH"
-. "$SOURCE_DIR/versions.bash"
 
 print_help() {
   cat <<-"EOF"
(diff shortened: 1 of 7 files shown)


bin/dfx-canister-url
```diff
@@ -2,7 +2,6 @@
 set -euo pipefail
 SOURCE_DIR="$(dirname "$(realpath "${BASH_SOURCE[0]}")")"
 PATH="$SOURCE_DIR:$PATH"
-. "$SOURCE_DIR/versions.bash"
```
bitdivine, line 5: Note: There are probably more files that import versions unnecessarily; I have not sought them all out.  I could do that for the next cleanup.  Started the PR but I'm out of time now.

dskloetd (approved):

PR #184, opened 2023-06-02
Raw canister urls

bitdivine:
# Motivation
Canister assets can also be obtained via raw URLs.

# Changes
- Add support for raw URLs.

# Tests
I have added tests for `--raw`.

### DIFF
--- a/bin/dfx-canister-url
+++ b/bin/dfx-canister-url
@@ -23,6 +23,7 @@ print_help() {
 	{ "canisters": {
 	    "my_canister": {
 	      "url": { "ic": "https://my-custom-url.com" },
+	      "raw_url": { "ic": "https://my-custom-raw-url.com" },
 	      ...
 	    }
 	  }
@@ -106,7 +107,7 @@ else
 fi
 
 # Not found?
-# Each network also has a pattern for canister URLs of the form: https://CANISTER_ID.NETWORK_URL
+# Each network also has a pattern for canister URLs of the form: https://CANISTER_ID.raw?.NETWORK_URL
 canister_id() {
   dfx canister id --network "$DFX_NETWORK" "$DFX_CANISTER"
 }
@@ -117,7 +118,7 @@ inject_canister_id() {
   static)
     local CANISTER_ID
     CANISTER_ID="$(canister_id)"
-    printf "%s\n" "$root_url" | sed -E "s,^(https?://)?,&${CANISTER_ID}.,g"
+    printf "%s\n" "$root_url" | sed -E "s,^(https?://)?,&${CANISTER_ID}${DFX_RAW:+.raw}.,g"
     ;;
   *) printf "%s\n" "$root_url" ;;
   esac
--- a/bin/dfx-canister-url
+++ b/bin/dfx-canister-url
@@ -97,7 +97,9 @@ while ! test -e dfx.json; do
   }
 done
 
-# Each network has a pattern for canister URLs of the form: https://CANISTER_ID.raw?.NETWORK_URL
+# Each network has a pattern for canister URLs of the forms:
+# - https://CANISTER_ID.NETWORK_URL
+# - https://CANISTER_ID.raw.NETWORK_URL
 canister_id() {
   dfx canister id --network "$DFX_NETWORK" "$DFX_CANISTER" 2>/dev/null || echo "$DFX_CANISTER"
 }
--- a/bin/dfx-canister-url
+++ b/bin/dfx-canister-url
@@ -64,7 +64,7 @@ URL_TYPES=(static api)
 clap.define short=n long=network desc="The dfx network to use" variable=DFX_NETWORK default="local"
 clap.define short=c long=canister desc="The canister name" variable=DFX_CANISTER
 clap.define short=t long=type desc="The canister type (options: ${URL_TYPES[*]})" variable=DFX_URL_TYPE default=static
-clap.define short=r long=raw desc="Provide a raw URL" variable=DFX_RAW nargs=0 default=""
+clap.define short=r long=raw desc="Provide a raw static URL" variable=DFX_RAW nargs=0 default=""
 clap.define short=o long=open desc="Open the URL in a browser" variable=DFX_OPEN nargs=0
 # Source the output file ----------------------------------------------------------
 source "$(clap.build)"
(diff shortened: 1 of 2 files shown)


bin/dfx-canister-url
```diff
@@ -63,6 +64,7 @@ URL_TYPES=(static api)
 clap.define short=n long=network desc="The dfx network to use" variable=DFX_NETWORK default="local"
 clap.define short=c long=canister desc="The canister name" variable=DFX_CANISTER
 clap.define short=t long=type desc="The canister type (options: ${URL_TYPES[*]})" variable=DFX_URL_TYPE default=static
+clap.define short=r long=raw desc="Provide a raw URL" variable=DFX_RAW nargs=0 default=""
```
bitdivine, line 67: Note:  It _may_ be better to have raw as a third type of URL.  Semantically, which makes more sense?  I'm not sure.

dskloetd: Is it possible to have a raw API URL that's different from the corresponding raw static URL or does that not make any sense?

I think the answer to this question determines whether it's a separate type of not.

bin/dfx-canister-url
```diff
@@ -97,13 +99,15 @@ done
 
 # Some canisters have custom URLs, such as identity.ic0.app.
 if [[ "${DFX_URL_TYPE:-}" == "static" ]]; then
-  url="$(jq -r '.canisters[env.DFX_CANISTER].url[env.DFX_NETWORK] // ""' dfx.json)"
+  key="${DFX_RAW:+raw_url}"
+  key="${key:-url}"
```
dskloetd, line 103: To me these 2 lines are (if I even understood them correctly) a very unreadable way of writing
```
if [ "$DFX_RAW" ]; then
  key="raw_url"
else
  key="url"
fi
```
Even something like this (assuming that it works) would be a lot more readable to me:
```
[ "$DFX_RAW" ] && key="raw_url" || key="url"
```

bin/dfx-canister-url
```diff
@@ -97,13 +99,15 @@ done
...
+  url="$(k="$key" jq -r '.canisters[env.DFX_CANISTER][env.k][env.DFX_NETWORK] // ""' dfx.json)"
 else
   url=""
 fi
 
 # Not found?
-# Each network also has a pattern for canister URLs of the form: https://CANISTER_ID.NETWORK_URL
+# Each network also has a pattern for canister URLs of the form: https://CANISTER_ID.raw?.NETWORK_URL
```
dskloetd, line 110: Is that `?` there correct? There is no `?` in a raw URL is there?

bin/dfx-canister-url.test
```diff
@@ -112,12 +112,13 @@ print_title() {
...
-  export ACTUAL_POSITIONAL_URL ACTUAL_FLAG_URL ACTUAL_POSITIONAL_URL_WITH_EXPLICIT_NETWORK ACTUAL_FLAG_URL_WITH_EXPLICIT_NETWORK
+  export ACTUAL_POSITIONAL_URL ACTUAL_FLAG_URL ACTUAL_POSITIONAL_URL_WITH_EXPLICIT_NETWORK ACTUAL_FLAG_URL_WITH_EXPLICIT_NETWORK ACTUAL_POSITIONAL_URL_WITH_RAW
   ACTUAL_POSITIONAL_URL="$(dfx-canister-url smiley_dapp_assets)"
   ACTUAL_FLAG_URL="$(dfx-canister-url --canister smiley_dapp_assets)"
   ACTUAL_POSITIONAL_URL_WITH_EXPLICIT_NETWORK="$(dfx-canister-url --network local smiley_dapp_assets)"
   ACTUAL_FLAG_URL_WITH_EXPLICIT_NETWORK="$(dfx-canister-url --network local --canister smiley_dapp_assets)"
-  for actual_varname in ACTUAL_POSITIONAL_URL ACTUAL_FLAG_URL ACTUAL_POSITIONAL_URL_WITH_EXPLICIT_NETWORK ACTUAL_FLAG_URL_WITH_EXPLICIT_NETWORK; do
+  ACTUAL_POSITIONAL_URL_WITH_RAW="$(dfx-canister-url --raw smiley_dapp_assets)"
+  for actual_varname in ACTUAL_POSITIONAL_URL ACTUAL_FLAG_URL ACTUAL_POSITIONAL_URL_WITH_EXPLICIT_NETWORK ACTUAL_FLAG_URL_WITH_EXPLICIT_NETWORK ACTUAL_POSITIONAL_URL_WITH_RAW; do
     [[ "$EXPECTED_URL" == "${!actual_varname}" ]] || {
```
dskloetd, line 122: I know this logic was here before but I'm confused. How can all of these have the same expected URL?
I'm thinking a long list of explicit calls with explicit expected URLs might be easier to follow.

bin/dfx-canister-url.test
```diff
@@ -207,6 +231,28 @@ print_title() {
     } >&2
   done
 )
+(
+  print_title "The mainnet API URLs should ignore the raw flag"
```
dskloetd, line 235: Wouldn't it be better for the script to fail if you specify `--raw` for API URLs, if that isn't meaningful?

bitdivine: I agree.  Done.

bin/dfx-canister-url
bitdivine: Discussed - raw API URLs are not a thing.  I have added a test "There are no raw API URLs" and updated the code accordingly.

bin/dfx-canister-url.test
bitdivine: This comment was clobbered but I can answer here.

> How can all of these have the same expected URL?

Because these commands are all asking for the same thing.  Or can you see two commands that should be giving different outputs?

bin/dfx-canister-url
```diff
@@ -95,17 +92,9 @@ while ! test -e dfx.json; do
...
-# Some canisters have custom URLs, such as identity.ic0.app.
-if [[ "${DFX_URL_TYPE:-}" == "static" ]]; then
-  url="$(jq -r '.canisters[env.DFX_CANISTER].url[env.DFX_NETWORK] // ""' dfx.json)"
-else
-  url=""
-fi
-
-# Not found?
-# Each network also has a pattern for canister URLs of the form: https://CANISTER_ID.NETWORK_URL
+# Each network has a pattern for canister URLs of the form: https://CANISTER_ID.raw?.NETWORK_URL
```
dskloetd, line 100: So the `?` is supposed to say `.raw` is optional? That wasn't clear to me.
It might be better to specify 2 full patterns, or put `.raw` in parentheses.

dskloetd (approved):

bin/dfx-canister-url
bitdivine: I have provided two patterns.

PR #185, opened 2023-06-03
Bump dfx to 0.14.1

bitdivine:
# Motivation
* We would like to deploy nns-dapp to app subnets.
* To make this convenient, we would like to let the controller upload a new frontend without re-compiling the backend with a new hard-wired whitelist.
* The `is_controller()` command is only supported in relatively recent replicas but dfx v0.13 uses a fairly ancient replica.
* So the version of dfx installed by snsdemo needs to be bumped to 0.14.1 or later.
* Then the version of dfx in nns-dapp can be bumped
* Then we can use `is_controller()` in CI
* Then we can merge the "deploy to an app subnet" code to th enns-dapp main branch.
* Then we can deploy release-candidates to app subnets.

# Changes
- Bump dfx to version 0.14.1
- Bump the macos version to v12.  (Currently available is 11,12,13.  Some tools such as the latest ic-wasm are known not to work with v11 and v11 has already been dropped by the nns-dapp repo).

# Tests
- See CI


peterpeterparker (approved):
Rubber stamp

PR #187, opened 2023-06-05
Update link used to find latest IC commit

bitdivine:
# Motivation
The link used to install ic-admin was updated in nns-dapp.  Checking whether the link needs to be updated here as well showed that the install link is already the same as in nns-dapp, but the link used to check the latest IC commit is an old one.

# Changes
- Update the link used to check whether an IC commit release includes ic-admin.

# Tests
The link is already used when installing.


mstrasinskis (approved):
Thank you!

dskloetd (approved):

PR #188, opened 2023-06-05
Bump macos to v12

bitdivine:
# Motivation
The current macos teleases available on GitHub are v11, v12 and v13.  We currently test snsdemo on v11.  The desire to update is also prompted by the fact that v11 no longer works with some tools, such as the latest release of ic-wasm.

Note: we can update to v12 or v13.  Either would work fine, as far as I know.

# Changes
- Update the macos version we test against from v11 to v12.

# Tests
See CI


peterpeterparker (approved):
Rubber stamp

PR #190, opened 2023-06-05
Actually bump dfx to v 0.14.1

bitdivine:
# Motivation
* We would like to deploy nns-dapp to app subnets.
* To make this convenient, we would like to let the controller upload a new frontend without re-compiling the backend with a new hard-wired whitelist.
* The `is_controller()` command is only supported in relatively recent replicas but dfx v0.13 uses a fairly ancient replica.
* So the version of dfx installed by snsdemo needs to be bumped to 0.14.1 or later.
* Then the version of dfx in nns-dapp can be bumped
* Then we can use `is_controller()` in CI
* Then we can merge the "deploy to an app subnet" code to the nns-dapp main branch.
* Then we can deploy release-candidates to app subnets.

# Changes
- Bump dfx to version 0.14.1
- Update string in test to match the new dfx.
- Mark `internet_identity` as not remote before deploying, as is now required.

# Tests
- See CI


dfx.json.original
bitdivine, line 1: Both need to be changed, or else the change is wiped out.

peterpeterparker (approved):
Rubber stamp

PR #191, opened 2023-06-06
snsdemo: Wait for aggregator

dskloetd:
# Motivation

For a snapshot to be useful out of the box, the sns_aggregator should be able to provide info on all existing SNSs.

# Changes

1. Update the used sns_aggregator version to one that has the candid interface in its metadata.
2. Change the candid placeholder in `dfx.json` to an actual candid path.
3. Add `bin/dfx-sns-aggregator-wait` which increases the speed of the aggregator, waits for a given number of SNSs to be loaded and then restores the aggregator config.
4. Call `bin/dfx-sns-aggregator-wait` from `bin/dfx-snapshot-stock-make`

# Tested

1. Ran `bin/dfx-snapshot-stock-make` manually (after `git clean -df` and `git checkout dfx.json`) and inspected the logs.
2. Add a test to `bin/dfx-stock-test`
3. CI

### DIFF
--- a/bin/dfx-sns-aggregator-wait
+++ b/bin/dfx-sns-aggregator-wait
@@ -82,10 +82,15 @@ wait_for_sns_count() {
   : Wait for aggregator to get all SNSs
   for ((try = 300; try > 0; try--)); do
     num_sns="$(get_sns_count)"
-    (("$num_sns" < "$EXPECTED_NUM_SNS")) || break
+    if (("$num_sns" == "$EXPECTED_NUM_SNS")); then
+      return
+    fi
     printf "\r #SNS: % 4d   Tries remaining: %4d" "$num_sns" "$try"
     sleep 2
   done
+  echo
+  echo "Failed to get $EXPECTED_NUM_SNS SNSs"
+  exit 1
 }
 
 save_aggregator_config
--- a/bin/dfx-sns-aggregator-wait
+++ b/bin/dfx-sns-aggregator-wait
@@ -1,0 +1,96 @@
+#!/usr/bin/env bash
+set -euo pipefail
+SOURCE_DIR="$(dirname "$(realpath "${BASH_SOURCE[0]}")")"
+PATH="$SOURCE_DIR:$PATH"
+
+print_help() {
+  cat <<-EOF
+
+	Waits for a given number of SNSs to be available in the sns_aggregator.
+	EOF
+}
+
+# Source the clap.bash file ---------------------------------------------------
+source "$SOURCE_DIR/clap.bash"
+# Define options
+clap.define short=n long=num_sns desc="The number of SNSs to wait for" variable=EXPECTED_NUM_SNS default="1"
+# Source the output file ----------------------------------------------------------
+source "$(clap.build)"
+
+: Make sure the candid is available and of the right version.
+DID_FILE="$(jq -r .canisters.sns_aggregator.candid dfx.json)"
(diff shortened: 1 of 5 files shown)


bin/dfx-sns-aggregator-wait
```diff
@@ -0,0 +1,96 @@
...
+  echo "Waiting for $EXPECTED_NUM_SNS SNSs"
+  page_size=10
+  required_page_count=$(((EXPECTED_NUM_SNS + page_size - 1) / page_size))
+
+  : Wait for aggregator to get all SNSs
+  for ((try = 300; try > 0; try--)); do
+    num_sns="$(get_sns_count)"
+    (("$num_sns" < "$EXPECTED_NUM_SNS")) || break
+    printf "\r #SNS: % 4d   Tries remaining: %4d" "$num_sns" "$try"
+    sleep 2
```
bitdivine, line 87: Should the command return a non-zero value or otherwise communicate if the condition is not met by the time of the timeout?

bin/dfx-software-sns-aggregator-install
```diff
@@ -54,7 +54,7 @@ test -e "$WASM_FILE" || {
 jq -e '.canisters.sns_aggregator' dfx.json >/dev/null || {
   echo "Adding sns_aggregator to dfx.json..."
   # TODO: Include API declaration .did files in release, so that we can get the candid.
-  jq -s '.[0] * .[1]' <(echo '{"canisters": { "sns_aggregator": { "type": "custom", "candid": "placeholder", "wasm": "'"${WASM_FILE}"'", "build": "true" }}}') dfx.json | sponge dfx.json
+  jq -s '.[0] * .[1]' <(echo '{"canisters": { "sns_aggregator": { "type": "custom", "candid": "candid/sns_aggregator.did", "wasm": "'"${WASM_FILE}"'", "build": "true" }}}') dfx.json | sponge dfx.json
```
bitdivine, line 57: How is the aggregator.did populated?

bin/dfx-sns-aggregator-wait
bitdivine, line 1: It would be nice to know that this works as advertised.  I see that you tested manually. Can you make an automated test as well please?  It doesn't have to be in this PR though.

bitdivine (approved):

bin/dfx-stock-test
```diff
@@ -28,7 +28,7 @@ export DFX_NETWORK
 
 (
   print_title "sns_aggregator should be installed"
-  index_html="$(mktemp --suffix .html index_XXXX)"
```
dskloetd, line 31: `--suffix` doesn't work on Mac and also seems to be unnecessary.

bin/dfx-software-sns-aggregator-install
dskloetd: Lines 20-22 in `bin/dfx-sns-aggregator-wait`:
```
: Make sure the candid is available and of the right version.
DID_FILE="$(jq -r .canisters.sns_aggregator.candid dfx.json)"
dfx canister metadata sns_aggregator candid:service >"$DID_FILE"
```

bin/dfx-sns-aggregator-wait
dskloetd: I updated `bin/dfx-stock-test`, which is run from `.github/workflows/snapshot.yml`

bin/dfx-sns-aggregator-wait
dskloetd: Good catch. Done.

PR #193, opened 2023-06-07
Test clap across many platforms

bitdivine:


bitdivine:
Deprioritized.  We decided to require bash.

PR #196, opened 2023-06-13
Update dfx nns import

bitdivine:
# Motivation
Now that we have dfx v 0.14.1 we have an improved `dfx nns import`.  It is better but not perfect so we need to update the polyfill, we can't discard it.

# Changes
- Discard the old polyfill
- Use the upstream `dfx nns import` as before, but with support for networks other than local.
- Replace the internet_identity did file with the one for the release we are using.

# Notes
## Intermittent test failure on "latest"
The location of the nns lifeline .did file seems to have changed in the ic repo.  With that change, `dfx nns import` breaks.  As such, this code will have to be revised soon.
* The "latest" test failed on Ubutu and Mac, failing to get https://raw.githubusercontent.com/dfinity/ic/ea6ab1dabffcfb7b901db3ec39002e972ebf69bc/rs/nns/handlers/lifeline/lifeline.did
* Looking in the ic repo, the .did file is now at: rs/nns/handlers/lifeline/impl/lifeline.did
* However after another push CI is green on "latest".  Interesting - has the location changed back??
## The future of dfx nns import
* `dfx deps` is a new command in dfx for handling canisters that have already been deployed.  It may potentially supersede `dfx nns import`; we would have to declare the NNS canisters as `"type": "pull"`, provide the canister IDs and then pull them from mainnet.  This functionality is however unstable so it is probably not a good idea to use it yet.
* `dfx deps` works well for replicating one network in another, it does not help when installing a specific commit of a project on a testnet.  It looks as if we may need to bring back an updated import command, that gets remotes and .did definitions from a  remote repo.


bin/dfx-ii-create-dummy-anchor
```diff
@@ -28,18 +28,15 @@ clean_up() {
 
 trap clean_up EXIT
 
-DECLARATIONS_DIR="$(realpath "$SOURCE_DIR/../declarations")"
-II_DID_FILE="$DECLARATIONS_DIR/internet_identity.did"
```
bitdivine, line 32: This is not where the upstream `dfx nns import` places the II .did file.  If the .did file location is unspecified, it will be picked up from `dfx.json` so let's delete this variable.  Perhaps in future all we will need is the II canister ID; we can get the .did from the deployed canister metadata, if we don't have it already.

bin/dfx-network-deploy
```diff
@@ -49,7 +49,7 @@ if [[ "$DFX_NETWORK" == "local" ]]; then
 
   dfx start --clean --background
   dfx-nns-install --ic_commit "$DFX_IC_COMMIT"
-  dfx-nns-import
+  dfx-nns-import --network "$DFX_NETWORK"
```
bitdivine, line 52: The effect of remote entries in dfx.json has changed in dfx v 0.14.1; e.g. __default remote IDs were previously a last resort and did not prevent deploying a canister.  Now it does, so adding a __default remote as before can cause deployments to break.  It seems best to be very specific about for which network canister IDs are wished.  Perhaps this is over-cautious and we can relax this but I would rather limit the scope until we have a clearer picture of how canister IDs will be managed in future.

bin/dfx-nns-import
bitdivine, line 1: This is the central change in this PR:
- Update the call to dfx nns import.
- Fix the result in any ways that it is deficient.

* dfx.json is populated more completely now, by `dfx nns import` but we still need to provide some canister IDs.
* The II did file is provided now but is not up to date with the version we deploy.  Rather than use a hard-wired .did file as before, this code now gets the .did file for the current II version.

PR #197, opened 2023-06-21
Use .wasm.gz

bitdivine:
# Motivation
The nns-dapp release WASMs are now called `.wasm.gz`.  The old suffix is still present but doesn't work for deploying with dfx v 0.14.1.

# Changes
- Update the nns-dapp and sns_aggregator release URLs and the paths of these files elsewhere in the repo.

# Tests
Existing CI should suffice, as this adds no new features or functionality.  Existing tests have been updated where needed.


lmuntaner (approved):
Thanks!

PR #200, opened 2023-06-26
Retry the neuron creation

bitdivine:
# Motivation
The neuron creation fails fairly often due to the network returning 503's.

# Changes
- Add a retry.

# Tests
- See CI
- I have deployed a testnet with this PR, albeit without command line arguments.

### DIFF
--- a/bin/dfx-neuron-create
+++ b/bin/dfx-neuron-create
@@ -7,6 +7,8 @@ source "$SOURCE_DIR/clap.bash"
 # Define options
 clap.define short=s long=icp desc="The amount of ICP to icp" variable=ICP default=1
 clap.define short=i long=DFX_IDENTITY desc="The dfx identity to use" variable=DFX_IDENTITY default="$(dfx identity whoami)"
+clap.define long=retries desc="The number of retries" variable=DFX_RETRIES default="20"
+clap.define long=retry_interval desc="The number of seconds between retries" variable=DFX_RETRY_SLEEP default="4"
 clap.define short=n long=network desc="The dfx network to use" variable=DFX_NETWORK default="local"
 clap.define short=V long=verbose desc="Print progress" variable=DFX_VERBOSE nargs=0
 # Source the output file ----------------------------------------------------------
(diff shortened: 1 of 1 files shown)


bin/dfx-neuron-create
```diff
@@ -7,56 +7,82 @@ source "$SOURCE_DIR/clap.bash"
...
-  exit 1
-} >&2
-NEURON_ID="$(echo "$NEURON_TEXT" | tr -cd '[:alnum:]')"
-test -n "${NEURON_ID:-}" || {
-  echo "Failed to get neuron ID from line: $NEURON_TEXT"
-  exit 1
-} >&2
+mk_neuron() {
+  (
+    set -euo pipefail
```
dskloetd, line 37: Does this have any effect?
When I tried adding this, it did *not* have the effect I wanted, which is:
1. When something inside `( ... )` fails, it doesn't continue running the block.
2. For the `( ... )` block to cause the outer script to fail when something inside it fails.

I tried this test script:
```#!/usr/bin/env bash
set -euo pipefail

DFX_RETRIES=5
DFX_RETRY_SLEEP=1

retry() {
  (
    set -eu
    for ((i = 0; i < DFX_RETRIES; i++)); do
      if "${@}"; then
        return
      else
        sleep "${DFX_RETRY_SLEEP}"
      fi
    done
    {
      echo "ERROR: Failed to run: $*"
      exit 1
    } >&2
  )
}

failing_func() {
  (
    set -euo pipefail
    echo before
    false
    echo after
  )
}

retry failing_func
```

And for me it outputs:
```
before
after
```

To get the intended effect, I have to replace `false` with `exit 1`.

bitdivine: Setting euo pipefail is just habit.  It's a good question, whether this does any good here.  Below the code does call exit explicitly when a required step fails, so I do believe the code works as intended.

dskloetd: On line 47:
```
IC_URL="$("$SOURCE_DIR/dfx-network-url" --network "$DFX_NETWORK")"
```
Previously, if the call to `dfx-network-url` would fail, the entire script would fail.
But now, if line 47 fails, it just continues with line 48.
This is surprising enough given the `set -euo pipefail` on line 2 but even more surprising given the `set -euo pipefail` on line 37.

If you want to be sure the script works the same as before, you'd have to replace line 47 with something like
```
if ! IC_URL="$("$SOURCE_DIR/dfx-network-url" --network "$DFX_NETWORK")"; then
  exit 1
fi
```
And similar for many other lines.
I'm not sure that's an improvement, but you should at least acknowledge it (perhaps with a comment on line 36) and it would be a good idea to go through the script and confirm that it fails when it should.

dskloetd (approved):

bitdivine:
Outdated

PR #204, opened 2023-07-04
Bump IC commit

bitdivine:
# Motivation
The IC commit is a month old. Patch files need to be updated for custom IC builds and we might as well update them to the latest versions.

# Changes
- Update the IC commit to the latest published version, which is on 3 July 2023.
- Update the code that parses the maximum amount a user is allowed to contribute.

# Tests
See CI


dskloetd:
It doesn't seem to work out of the box? Let me know when I should have a look.

dskloetd (approved):

PR #210, opened 2023-08-17
Stop replica when dfx-snapshot-stock-make fails

dskloetd:
# Motivation

Multiple people have been confused by the replica continuing to log after `dfx-snapshot-stock-make` fails.
This makes it hard to see if the script has finished and what the error was.

# Changes

1. Set a trap in `bin/dfx-snapshot-stock-make` to stop the replica and to explain that the script fails.
2. Remove the trap once the replica is stopped normally, otherwise it would still trigger when the script exits successfully.

# Tested

I introduced a type in another script and verified that the replica was stopped and the error message displayed.
Then I removed the type and verified that the snapshot is made successfully without error message.


mstrasinskis (approved):
Thanks!

PR #211, opened 2023-08-25
Action to check for dfx updates

bitdivine:
# Motivation
We would like to automate chores, such as updating dfx, IC commit and so on.  And we now have the requisite permissions to do so in the snsdemo.

Note: Updates create a PR, so a human can decide whether a  given update is appropriate.

Let's start with dfx.

# Changes
- Add a github workflow that checks for dfx updates and, on finding one, creates a corresponding PR.

# Testing
- This action generated this PR: https://github.com/dfinity/snsdemo/pull/212

### DIFF
--- a/.github/actions/slack/README.md
+++ b/.github/actions/slack/README.md
@@ -1,0 +1,21 @@
+# Slack
+
+A GitHub Action for sending slack notifications. For more information, see slack's [documentation](https://api.slack.com/messaging/webhooks). In particular, note that the channel and workspace are set by the webhook URL, not directly by this action.
+
+## Usage
+
+```yaml
+name: My Action
+
+on:
+  push:
+
+jobs:
+  share-love:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: ./.github/actions/slack
+        with:
+          WEBHOOK_URL: ${{ secrets.SLACK_WEBHOOK_URL }}
+          MESSAGE: "I love you guys."
+```
--- a/.github/workflows/update-dfx.yml
+++ b/.github/workflows/update-dfx.yml
@@ -6,8 +6,10 @@ on:
     # check for new dfx versions weekly
     - cron: '30 3 * * FRI'
   workflow_dispatch:
+  # Provide an option to run this manually.
   push:
     branches:
+      # Development branch for this workflow:
       - "update-dfx"
 jobs:
   rust-update:
@@ -44,7 +46,7 @@ jobs:
           commit-message: Update dfx version
           committer: GitHub <noreply@github.com>
           author: gix-bot <gix-bot@users.noreply.github.com>
-          branch: bot-rust-update
+          branch: bot-dfx-update
           delete-branch: true
           title: 'Update dfx version'
           # Since the this is a scheduled job, a failure won't be shown on any
--- a/.github/workflows/update-dfx.yml
+++ b/.github/workflows/update-dfx.yml
@@ -28,8 +28,10 @@ jobs:
 
           if [ "$current_dfx_version" != "$latest_dfx_version" ]
           then
-            version="$latest_dfx_version" jq '.dfx=(env.version)' dfx.json > dfx-new.json
-            mv dfx-new.json dfx.json
+            for file in dfx.json dfx.json.original ; do
+              version="$latest_dfx_version" jq '.dfx=(env.version)' "$file" > "$file.new"
+              mv "$file.new" "$file"
+            done
             echo An update is available for dfx.
             git diff
             echo "updated=1" >> "$GITHUB_OUTPUT"
--- a/.github/workflows/update-dfx.yml
+++ b/.github/workflows/update-dfx.yml
@@ -43,7 +43,7 @@ jobs:
         with:
           token: ${{ secrets.GIX_SNSDEMO_BOT_GH_TOKEN }}
           base: main
-          add-paths: ./dfx.json
+          add-paths: dfx.json
           commit-message: Update dfx version
           committer: GitHub <noreply@github.com>
           author: gix-bot <gix-bot@users.noreply.github.com>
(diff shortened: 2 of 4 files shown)


.github/workflows/update-dfx.yml
```diff
@@ -0,0 +1,64 @@
...
+          delete-branch: true
+          title: 'Update dfx version'
+          # Since the this is a scheduled job, a failure won't be shown on any
+          # PR status. To notify the team, we send a message to our Slack channel on failure.
+      - name: Notify Slack on failure
+        uses: ./.github/actions/slack
+        if: ${{ failure() }}
+        with:
+          WEBHOOK_URL: ${{ secrets.SLACK_WEBHOOK_URL }}
+          MESSAGE: "Dfx update failed"
```
dskloetd, line 64: `WEBHOOK_URL` and `MESSAGE` are passed to the GitHub action inputs `webhook_url` and `message`, correct?
Even though the inputs are apparently case insensitive (?) can we use the same case in both places to avoid confusion?
Especially since `WEBHOOK_URL` and `MESSAGE` are also used in the action, but as `env` rather than `inputs`.

.github/actions/slack/README.md
```diff
@@ -0,0 +1,21 @@
...
+  push:
+
+jobs:
+  share-love:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: ./.github/actions/slack
+        with:
+          WEBHOOK_URL: ${{ secrets.SLACK_WEBHOOK_URL }}
+          MESSAGE: "I love you guys."
```
dskloetd, line 20: Why do you pass the inputs in upper case, even though they are declared as lower case?

.github/workflows/update-dfx.yml
```diff
@@ -0,0 +1,64 @@
...
+            echo An update is available for dfx.
+            git diff
+            echo "updated=1" >> "$GITHUB_OUTPUT"
+          else
+            echo "updated=0" >> "$GITHUB_OUTPUT"
+          fi
+        # If a newer dfx is available, create a PR.
+      - name: Create Pull Request
+        if: ${{ steps.update.outputs.updated == '1' }}
+        uses: peter-evans/create-pull-request@v4
```
dskloetd, line 44: Is it safe to give Mr. Peter Evens access to our repository like this? Has this been cleared with security?

.github/workflows/update-dfx.yml
```diff
@@ -0,0 +1,64 @@
...
+        # latest version's tag name is the version.
+      - name: Check new dfx version
+        id: update
+        run: |
+          current_dfx_version="$(jq -r .dfx dfx.json)"
+          echo "Current dfx version: '$current_dfx_version'"
+          latest_dfx_version="$(curl -sSL https://sdk.dfinity.org/manifest.json | jq -r .tags.latest)"
+          echo "Latest dfx version: '$latest_dfx_version'"
+
+          if [ "$current_dfx_version" != "$latest_dfx_version" ]
```
dskloetd, line 29: If I recall correctly, we started using version `0.13.0` in nns-dapp before it was officially marked as the latest version. Should we check for `-lt` instead of `!=`?

.github/workflows/update-dfx.yml
bitdivine: Peter Evans is already all over dfinity repositories: https://github.com/search?q=org%3Adfinity%20peter-evans&type=code

.github/workflows/update-dfx.yml
bitdivine: I would not trust bash lt to compare semantic version numbers out of the box.

I believe that exceptional circumstances, such as using `0.14.1` when `0.14.0` is still marked as latest is best dealt with manually, by rejecting the PR.

.github/actions/slack/README.md
bitdivine: This is not a decision made here.  The action is copied directly from nns-dapp (which in turn was copied from II).  We should really publish it so that we don't have to copy it all over the place.  But there is a reason why the actions aren't shared internally:

It doesn't seem to be possible [to reference an action defined in another dfinity repo](https://github.com/actions/checkout/issues/95).  We could make a dedicated repo, called `dfinity/github-actions` or some such, and clone it in other repos before use.  I'm not sure that works but we can try.

.github/workflows/update-dfx.yml
bitdivine: I recommend discussing this with upstream.  This code replicates upstream: https://github.com/dfinity/internet-identity/blob/main/.github/workflows/update-dapps.yml  Having different conventions in different places makes it harder to compare code.  I would rather have consistency.

.github/actions/slack/README.md
dskloetd: Can you add a link to the original here in the file?

.github/workflows/update-dfx.yml
dskloetd: Next time you include copied code in a PR, can you mention in the PR description that it's copied and from where? Thanks.

bitdivine: Sure thing.  If we spend time on this, I would like to have a hard look at factoring out the action so that the copying is unnecessary.

.github/actions/slack/README.md
bitdivine: Sure thing.

bitdivine: Done.  I have added the source as a `SOURCE.md` so that importing changes doesn't require adding the line.

dskloetd: I don't see any `SOURCE.md`.

bitdivine: I was checking something locally.  Pushed.

dskloetd (approved):

PR #218, opened 2023-09-08
Create more SNS token for the faucet

dskloetd:
# Motivation

When we create an snsdemo snapshot, we send tokens to the faucet address so you can use the "Get Tokens" button to get tokens.
Currently we only create 3141 tokens and only half of that can be disbursed.

# Changes

1. Create 10'003'141 tokens instead of 3'141 so we can use the faucet for longer before it runs dry.

# Tested

Manually ran the script and tried to get 4'000'000 SNS tokens.

### DIFF
--- a/bin/dfx-sns-sale-propose
+++ b/bin/dfx-sns-sale-propose
@@ -21,7 +21,7 @@ export DFX_NEURON_ID DFX_IDENTITY_PEM DFX_NNS_URL
 # The due date of the sale must be greater than 1 day at the time of NNS proposal execution
 SWAP_DUE_TIMESTAMP="$(perl -e 'print time() + 90 * 24 * 3600')"
 SNS_SWAP="$(dfx canister id sns_swap --network "$DFX_NETWORK")"
-SNS_TOKENS_E8s=314100000000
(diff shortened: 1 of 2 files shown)


lmuntaner (approved):
Thanks!

Btw, worth taking into account this one: https://dfinity.slack.com/archives/C0175CHJ0P6/p1694217346613589

It might simplify how we get the tokens, right?

dskloetd:
I wasn't aware of that call when I made this change but I'm also a bit torn about it.
It would certainly have been simpler if it existed from the beginning but using it now will definitely be a bigger change than this PR.
And should we use it in snsdemo instead of disbursing a neuron, or in the NNS dapp when we actually request the tokens?
Anyway, I don't know when we'll justify actually making either of those changes. We might just end up making that call manually occasionally.

dskloetd:
Not yet: https://github.com/dfinity/snsdemo/pull/220

Btw, strange that it's the governance canister and not the ledger canister minting the tokens...

lmuntaner:
> Not yet: #220
> 
> Btw, strange that it's the governance canister and not the ledger canister minting the tokens...

Yes 😂

PR #232, opened 2023-10-06
Install KYT as part of ckBTC

dskloetd:
# Motivation

We want to be able to test ckBTC locally and on CI.
The ckBTC minter calls the KYT (know your transaction?) canister to check if the tokens aren't "tainted".
So we need a KYT canister installed.

# Changes

1. In `bin/dfx-ckbtc-import`, add entries to `dfx.json`, download WASM and download candid for kyt canister. Same as is done for the minter, ledger and index canister.
2. In `bin/dfx-ckbtc-deploy`, create kyt canister, deploy kyt, and set an API key on the canister.
3. Drive-by: Include `index` in the `check_ckbtc_deployment()`. It seems like it might have been forgotten and it doesn't hurt to check.

This is necessary but not sufficient to test ckBTC. We still need to install a (mock) bitcoin canister in a future PR.

# Tested

1. Updated `bin/dfx-ckbtc-deploy.test` in a way consistent with the existing code.
2. Manually ran a snapshot, install a mock bitcoin canister on top of it and tested that I was able to (mock) send BTC.

### DIFF
--- a/bin/dfx-ckbtc-deploy
+++ b/bin/dfx-ckbtc-deploy
@@ -35,14 +35,32 @@ source "$(clap.build)"
 } >&2
 
 deploy_ckbtc() {
+  # We create some canisters ahead of time because we need to know their IDs
+  # before deploying. Remaining canisters will be created when deployed.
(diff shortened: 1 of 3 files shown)


lmuntaner (approved):
Thanks!

lmuntaner (approved):
Thanks!

PR #242, opened 2023-10-18
Don't skip downloading ckBTC wasms.

dskloetd:
# Motivation

We don't know the version of the existing wasms and we should make sure the installed ckBTC wasms match the specified IC commit.

# Changes

1. Remove the code that skips downloading ckBTC wasms if the file already exists.
2. Write output from `curl` with `>` instead of `-o` to directly overwrite existing files.

# Tests

Manually created a stock snapshot in a directory that had pre-existing wasms.


lmuntaner (approved):
Thanks!

PR #243, opened 2023-10-18
Make number of parallel SNSes configurable

dskloetd:
# Motivation

When making changes to the snapshot setup or if you just don't need 12 SNSes, it can be useful to configure the number of SNSes included in the snapshot to make it faster.

# Changes

1. Add flag `--parallel_sns_count` to `bin/dfx-snapshot-stock-make` and `bin/dfx-stock-deploy`. Note that `dfx-sns-demo` always creates 2 SNSes so the flag controls the number of additional SNSes (which are created in parallel) created after that.
2. Unrelated-ish: Don't hide errors when trying to count the number of SNSes.

# Tested

Manually tested with `--parallel_sns_count 3` and `--parallel_sns_count 0`.


lmuntaner (approved):
Thanks!

PR #244, opened 2023-10-19
Update IC commit

sa-github-api:
Automated changes by [create-pull-request](https://github.com/peter-evans/create-pull-request) GitHub action

PR #247, opened 2023-10-19
Chore: upgrade IC commit

lmuntaner:
Upgrading IC Commit.

Checking why #244 is stuck.


lmuntaner:
Not needed

dskloetd:
Please don't ignore things like that. Because then when you are on vacation, we get stuck with it: https://dfinity.slack.com/archives/CL7Q2RXUM/p1697014801586649

bitdivine:
Believe me, they are not ignored.

dskloetd:
In that case, please update us about what's being done about the cla issue.

PR #255, opened 2023-10-30
Bump internet identity version

dskloetd:
# Motivation

The current used version of Internet Identity is from April. It has a bug where it tries to connect to mainnet when it doesn't recognize the FE URL. This prevents us from using it on a DevEnv.

# Changes

1. Update Internet Identity to the newest version from 2023-10-27.
2. Fix `bin/dfx-ii-create-dummy-anchor` to pass `null` to the extra optional argument of the `register` method, that was added since the previous version.

# Tested

I created a snapshot manually and prepared https://github.com/dfinity/nns-dapp/pull/3637 to update the E2E tests to the new ii.


mstrasinskis (approved):
Thank you!

PR #259, opened 2023-11-03
Update dfx version

sa-github-api:
Automated changes by [create-pull-request](https://github.com/peter-evans/create-pull-request) GitHub action


dskloetd:
I'm not a fan of this one. Having different version of dfx in different repos is a bit of a pain. And it doesn't seem very import to have the newest version. Maybe we can go back to doing this as needed?

bitdivine:
> Having different version of dfx in different repos is a bit of a pain.
Which repos are you thinking of?

> And it doesn't seem very import to have the newest version.
A weekly reminder that we still haven't switched to using extensions is reasonable.

PR #263, opened 2023-11-07
No wasm nm

bitdivine:
# Motivation
We use `wasm-nm` (the Wasm version of [`nm`](https://web.mit.edu/gnu/doc/html/binutils_3.html)) to geta  list of exported functions.

Unfortunately `wasm-nm` is no longer maintained and fails with the latest Wasm files that have new instructions.

`ic-wasm info` provides the same information, but not in a machine readable form.  Where we have switched to ic-wasm we use sed to extract the interesting information, at which point the output is practically in the standard `nm` output format.

I propose to ask upstream to provide a machine readable output, preferably in the `nm` format, and in the meantime make a polyfill.

## Alternatives considered

Copy and paste the sed incantation into the one place we use `wasm-nm` in this repository.  I'd rather get a clean solution pushed upstream.


lmuntaner (approved):
Who is upstream in this case?

bitdivine:
Upstream ic `ic-wasm`.  I threw in [a PR](https://github.com/dfinity/ic-wasm/pull/52) to address this issue.

lmuntaner:
> Upstream ic `ic-wasm`. Update: I communicated with them and threw in [a PR](https://github.com/dfinity/ic-wasm/pull/52) to address this issue.

Thansk!

PR #264, opened 2023-11-07
feat: Installer action for binstall

bitdivine:
# Motivation
We would like to replace `wasm-nm`, which is unmaintained, with `ic-wasm`.  The latter is installed with `binstall` in the `nns-dapp` repo.  We propose to follow the same pattern, the only difference being where the binstall version is stored.

# Changes
- Add a GitHUb action that installs binstall.#263 

# Tests
The code is exercised heavily in `nns-dapp`, differing only in where `BINSTALL_VERSION` is sourced from.


.github/actions/install_binstall/action.yaml
```diff
@@ -0,0 +1,12 @@
+name: 'Installs binstall'
+description: |
+  Installs `cargo binstall` at the version specified in dfx.json
```
lmuntaner, line 3: The version is specified in `versions.bash`, not dfx.json

bitdivine: Thank you.  Good catch.  Done.

lmuntaner (approved):
Thanks!

PR #266, opened 2023-11-08
feat GIX-2044 (CI): Bot to update aggregator

bitdivine:
# Motivation
We would like to keep the SNS aggregator in demo snapshots reasonably up to date.

# Changes
- Add a bot that checks for a new SNS aggregator and, when it finds one, updates the pinned version.

# Tests
This code created this PR: https://github.com/dfinity/snsdemo/pull/265


.github/workflows/update-aggregator.yml
```diff
@@ -0,0 +1,74 @@
...
+        uses: peter-evans/create-pull-request@v4
+        with:
+          token: ${{ secrets.GIX_SNSDEMO_BOT_GH_TOKEN }}
+          base: main
+          add-paths: |
+            bin/versions.bash
+          commit-message: Update SNS Aggregator version
+          committer: GitHub <noreply@github.com>
+          author: gix-bot <gix-bot@users.noreply.github.com>
+          branch: bot-aggregator-update
```
dskloetd, line 54: Can we use a unique branch name so consecutive runs of the action don't interfere with each other?

lmuntaner: If possible and easy I agree with this.

So that we can use it in all the other workflows. Maybe add the date in the name of the branch if it's easier.

dskloetd: Date + time so you can trigger it as often as you want.

bitdivine: How many concurrent PRs, all updating the aggregator commit, do you want?

Note: You can trigger the workflow as often as you want, that is an independent issue.

bitdivine: Personally, I actively _do_ want the workflow to be able to update the PR with the latest commit, if a current PR has not been merged yet.  I actively do not think it makes sense to have multiple PRs all updating the aggregator version.

dskloetd: What is the benefit of overwriting an existing PR as opposed to creating a new PR?
Also, from memory, I don't think it actually supports updating an existing PR. I think it still creates a new PR but also confusingly messes with the old PR at the same time.
We can avoid confusion by always using a new branch.

lmuntaner: I also think that it created two PRs. That's why I prefer two branches as well. Then it's clear that they are two PRs.

Checking whether there is already an open PR with that branch might be harder.

bitdivine: > Also, from memory, I don't think it actually supports updating an existing PR. I think it still creates a new PR but also confusingly messes with the old PR at the same time.

Your memory deceives you.

When developing this bot, I ran the bot repeatedly.  It updated [one and only one PR](https://github.com/dfinity/snsdemo/pull/265).  You can check that the bot created only one PR if you look at [the list of recent PRs](https://github.com/dfinity/snsdemo/pulls?q=is%3Apr).  In that one PR, this is what the update looks like in the PR history:

Screenshot from 2023-11-08 15-18-46

GitHub _cannot_ have two simultaneous open PRs with the same source and destination branch.

bitdivine: > Then it's clear that they are two PRs.
Why do you want two PRs open at the same time, both updating the aggregator version?

> Checking whether there is already an open PR with that branch might be harder.
This is actually trivial.  If I want to find out whether there is an open PR generated by this bot:
```
gh pr list --base main --head bot-aggregator-update
```

dskloetd: Max,
A while back you created this with a bot: https://github.com/dfinity/nns-dapp/pull/3289
I left a lot of comments on it.
You then suggested we discuss in person which we did for a bit. But then you said you had time boxed the task and you would rather continue another time. So I left it.
Then suddenly, this other PR appears **from the same branch**: https://github.com/dfinity/nns-dapp/pull/3369
And it turned out you completely ignored all my comments and bypassed my review and merged a change that resulted in exactly the problems that I anticipated.

On slack you also wrote to me: "... At present, if the bot runs again, it wipes out the existing description.  This has happened to me ..."

So now we are both telling you that we prefer a unique branch.

bitdivine: > I left a lot of comments on it.

You left two comments in that PR.  Is that a lot?

> ...  time boxed the task

Yes, I time-box everything.

> Then suddenly, ....

Can you clarify?  Your description does not match the actual comments on file.

> So now we are both telling you that we prefer a unique branch.

The facts that you have cited in your reasoning have been false.  Agreed?

Of course, if you are not happy with how the maintenance is done, you are welcome to take over the epic.  I am not in the business of forcing my opinions down other people's throats.

bitdivine: On the contrary, I have read your responses and responded to them.  I do note that the bulk of your argument is based on factually incorrect statements.  (Two specific examples:  "I don't think it actually supports updating an existing PR.  I think it still creates a new PR but also confusingly messes with the old PR at the same time.")  If the premise of an argument turns out to be false, I think you will agree that the argument needs to be reconsidered, yes?

bitdivine: If you like, you can consider the actual situation and make a fresh argument on a solid basis, if you still believe in your conclusion.

bitdivine: Branch per PR  added by popular demand.

I do note that this may cause the workflow to become spammy, which may cause people to want to run it less frequently and so cause testing with old code.  We will see.

lmuntaner:
Outdated

PR #267, opened 2023-11-08
Check whether creating an anchor has succeeded

bitdivine:
# Motivation
If creating an anchor fails, subsequent steps are likely to fail as well so it is worth catching the error early.

Unfortunatley `dfx canister call` does not return with a non-zero exit code when the canister returns an error.

However we can easily and cheaply check whether the newly created anchor is indeed an integer.

# Changes
- Add a check that a newly created anchor is indeed an integer.


lmuntaner (approved):
Rubber stamp!

PR #268, opened 2023-11-09
Make II install mode variable

bitdivine:
# Motivation
Sometimes it is desirable to upgrade rather than reinstalling II

# Changes
- In the II install utility, make the install mode a variable while leaving the default as it is now.

# Tests
- The default behaviour is exercised in CI.
- I have used `--mode upgrade` when developing and it has worked fine.


lmuntaner (approved):
LGTM!

When did you need this?

bitdivine:
@lmuntaner  It would be used when a developer has created some identities but then want to check whether everything still works with an updated II.  I'm not sure it would have any use in CI but I found it useful when debugging the snapshot creation.  It seems worth keeping for the next debugging session.

lmuntaner:
> @lmuntaner It would be used when a developer has created some identities but then want to check whether everything still works with an updated II. I'm not sure it would have any use in CI but I found it useful when debugging the snapshot creation. It seems worth keeping for the next debugging session.

Thanks for the explanation!

PR #269, opened 2023-11-09
Create dfx entry if needed

bitdivine:
# Motivation
`dfx nns import` is supposed to create entries for all the NNS canisters in the local `dfx.json`.  Unfortunately with `dfx v0.15.1` the II candid is not installed and the canister is missing from `dfx.json`.  We replace the candid file with a newer version anyway, that is not a problem, but we _do_ need the entry in `dfx.json`.

Note: I tested the behaviour of `dfx nns import` for versions `0.14.4` and `0.15.1` in clean new projects (created with `dfx new PROJECT_NAME`).  This behaviour is in no way specific to the use in the snsdemo repository.

# Changes
- When updating the II candid, ensure that the path to the candid file is also listed in `dfx.json`.

# Tests
- For `dfx 0.14.4` see the CI for this PR.
- For `dfx 0.15.1` see the CI for the PR that updates dfx to `0.15.1`.

### DIFF
--- a/bin/dfx-nns-import
+++ b/bin/dfx-nns-import
@@ -45,9 +45,17 @@ export DFX_NETWORK
 # works as long as the .did files are not moved in the ic repo.
 DFX_IC_COMMIT="$DFX_IC_COMMIT" dfx nns import --network-mapping "$DFX_NETWORK=mainnet"
 
-# Update the II candid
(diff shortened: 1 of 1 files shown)


lmuntaner (approved):
If I understand correctly, this works for both dfx versions, right?

bitdivine:
@lmuntaner  Yes, this works for both `dfx` `0.14.4` (and presumably earlier) and `0.15.1`.  Thanks for the review.

PR #262, opened 2023-11-07
Dfx 15

bitdivine:
# Motivation
We would like to keep `dfx` up to date.

# Changes
- Update `dfx` from version `0.14.4` to `0.15.1`
- Install `dfx nns` as an extension, as it is no longer in `dfx` core.

# Tests
- See CI
- [Works with the nns-dapp.](https://github.com/dfinity/nns-dapp/pull/3716)

PR #270, opened 2023-11-09
Make the current user a controller of the II and nns-dapp

bitdivine:
# Motivation
In `dfx v0.14.4`, `dfx nns deploy` would create the `internet_identity` and `nns-dapp` canisters with the current user as the controller.  In `dfx v0.15.1` the anonymous user is the controller, which is a breaking change.

# Changes
- Check whether the current user is a controller and if not, use the anonymous user to add the current user as a controller.

# Tests
- For `dfx v0.14.4` see CI in the current PR.
- For `dfx v0.15.1` see [the PR that updates `dfx` to `v0.51.1`](https://github.com/dfinity/snsdemo/pull/262).


lmuntaner (approved):
LGTM! Thanks!

PR #272, opened 2023-11-10
Correct arguments to reelase action

bitdivine:
# Motivation
The "Create snapshot image" action emits this warning:

Screenshot from 2023-11-10 16-01-58

And, indeed, the release action doesn't need that argument.

# Changes
- Remove the `token` argument from the call to `./.github/actions/release`

# Tests
See CI - The snapshot action should have no such warning.


lmuntaner (approved):
Thanks!

PR #274, opened 2023-11-14
Fix dfx-nns-import

dskloetd:
# Motivation

In `dfx-nns-import` on [this line](https://github.com/dfinity/snsdemo/blob/24dc8ca57747853201895a496e9bceabf4fb4277/bin/dfx-nns-import#L55), `jq` waits for input.
The comment says an entry for `internet_identity` is added to `dfx.json` but that line doesn't do anything with `dfx.json`.
This code was added in https://github.com/dfinity/snsdemo/pull/269 and apparently not tested manually.
On CI the user input is apparently automatically closed so the script is able to continue.
It will then rely on the fallback on line 64 to still get internet_identity candid.

# Changes

Actually update `dfx.json` instead of waiting for user input.
`//` is the alternative operator: https://jqlang.github.io/jq/manual/#alternative-operator
So `//=` only assigns if there isn't already a value.


lmuntaner (approved):
LGTM! Thanks!

PR #277, opened 2023-11-20
Update SNS aggregator version to proposal-125509-agg

sa-github-api:
# Motivation
A new version of the SNS aggregator is available.  We would like to keep the snapshots up to date.

# Changes
* Updated the pinned default SNS aggregator version.

# Tests
  - Please see CI.

### DIFF
--- a/bin/versions.bash
+++ b/bin/versions.bash
@@ -2,7 +2,7 @@
 # shellcheck disable=SC2034 # Variables are expected to be unused in this file
 # Release from 2023-05-30 which includes geo restriction fields
 SNS_AGGREGATOR_RELEASE=proposal-125509-agg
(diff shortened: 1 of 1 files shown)


lmuntaner:
@bitdivine we should wait until after the release to be sure that it's the one in prod, right?

bitdivine:
If this is meant to be the one in prod (a perfectly reasonable suggestion) I would suggesting updating the bot to get the `aggregator-prod` tag by default and making a PR if the commit of that tag has changed.  What do you think?

lmuntaner:
> @lmuntaner If this is meant to be the one in prod (a perfectly reasonable suggestion) I would suggesting updating the bot to get the `aggregator-prod` tag by default and making a PR if the commit of that tag has changed. What do you think?

Makes sense to me!

lmuntaner (approved):
Good to go now!

PR #278, opened 2023-11-20
Fix: short form was repeated

bitdivine:
# Motivation
The flag `-l` was ambiguous in some `bin/dfx-software-.*-version` commands, with several options having the same short form.

There is a risk of "just changing the flag" but in a way that makes the flags inconsistent or confusing so I reviewed all these flags. This convention seems to make sense and is mostly consistent with what we already have:

* `-l --latest` The latest release.
* `-m --mainnet` The version deployed on mainnet.
* `-p --pinned` The version in bin/versions.bash
* `-v --version` A named release, such as "latest".  Enables passing through a version from one command to the next by name rather than with lots of different flags.

# Changes
* Change the short form for "pinned" to not conflict.
* Implement the above convention consistently.

# Tests
I have searched for usages of `dfx-software-.*-version` and found that all uses in `snsdemo` and `nns-dapp` use the long form, so changing the short flags will not affect CI.  It may affect command line usage - that is where I personally use short flags most frequently.

### DIFF
--- a/bin/dfx-software-didc-version
+++ b/bin/dfx-software-didc-version
@@ -16,6 +16,7 @@ source "$SOURCE_DIR/clap.bash"
 # Define options
 clap.define short=p long=pinned desc="Get the pinned version from versions.bash" variable=GET_PINNED nargs=0
 clap.define short=l long=latest desc="Get the latest version" variable=GET_LATEST nargs=0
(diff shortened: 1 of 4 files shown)


lmuntaner (approved):
Thanks!

PR #279, opened 2023-11-20
feat: Add option to get mainnet aggregator version

bitdivine:
# Motivation
We would like to be able to easily update the aggregator to the version in production.

# Changes
* Add a `--mainnet` option to the command that looks up aggregator versions.

# Tests
A test has been added that the command returns a proposal tag.

### DIFF
--- a/bin/dfx-software-sns-aggregator-version
+++ b/bin/dfx-software-sns-aggregator-version
@@ -20,16 +20,26 @@ source "$SOURCE_DIR/clap.bash"
 # Define options
 clap.define short=l long=pinned desc="Get the pinned version from versions.bash (default)" variable=GET_PINNED nargs=0
 clap.define short=l long=latest desc="Get the latest version" variable=GET_LATEST nargs=0
+clap.define short=m long=mainnet desc="Get the version on mainnet" variable=GET_MAINNET nargs=0
 # Source the output file ----------------------------------------------------------
 source "$(clap.build)"
 
 STRATEGY="${STRATEGY:-${GET_PINNED:+pinned}}"
 STRATEGY="${STRATEGY:-${GET_LATEST:+latest}}"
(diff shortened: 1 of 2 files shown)


bin/dfx-software-sns-aggregator-version
```diff
@@ -20,17 +20,27 @@ source "$SOURCE_DIR/clap.bash"
 # Define options
 clap.define short=l long=pinned desc="Get the pinned version from versions.bash (default)" variable=GET_PINNED nargs=0
 clap.define short=l long=latest desc="Get the latest version" variable=GET_LATEST nargs=0
```
bitdivine, line 22: Note: There is a bug above where the short form `-l` is repeated.  This error has been replicated across several commands so is fixed in a separate PR.

bin/dfx-software-sns-aggregator-version
```diff
@@ -20,17 +20,27 @@ source "$SOURCE_DIR/clap.bash"
...
 STRATEGY="${STRATEGY:-${GET_LATEST:+latest}}"
+STRATEGY="${STRATEGY:-${GET_MAINNET:+mainnet}}"
 STRATEGY="${STRATEGY:-pinned}" # Default - not guaranteed to be stable
 case "${STRATEGY}" in
 pinned)
   . "$SOURCE_DIR/versions.bash"
   echo "$SNS_AGGREGATOR_RELEASE"
   ;;
+mainnet)
+  prod_commit="$(git ls-remote git@github.com:dfinity/nns-dapp.git aggregator-prod | awk '{print $1}')"
```
bitdivine, line 37: Note: Alternatively we could get the commit with:
```
 dfx canister metadata 3r4gx-wqaaa-aaaaq-aaaia-cai git_commit_id --network ic
```
This might be better but assumes that `dfx` is installed.  We usually have `dfx` installed so it's not much of a hurdle.

bitdivine:
Outdated

PR #280, opened 2023-11-24
Update dfx version

sa-github-api:
Automated changes by [create-pull-request](https://github.com/peter-evans/create-pull-request) GitHub action


bitdivine:
@dskloetd How is progress on getting dfx 0.15.xx to work with devenv?

I'll reject upgrades like this until we have a working setup with devenv.

PR #282, opened 2023-11-27
Avoid using decimal comma on machines with different locale

dskloetd:
# Motivation

This error was reported:
```
/Users/daviddalbusco/.config/dfx/identity/snsdemo8/identity.pem transfer --amount 5001570,49997000 --fee 0,00001000 jg6qm-uw64t-m6ppo-oluwn-ogr5j-dc5pm-lgy2p-eh6px-hebcd-5v73i-nqe
error: Invalid value "5001570,49997000" for '--amount <AMOUNT>': Failed to parse tokens as unsigned integer
```
which seems to happen because a `awk` output a decimal comma instead of a decimal point.

This PR might fix it for the line where the error happened but I'm not sure (actually I would expect it) if there are other places that would have the same problem.


yhabib (approved):
Thanks

PR #283, opened 2023-11-27
Scripts to deploy generic token

dskloetd:
# Motivation

We want to deploy a ledger+index canister to test ckETH functionality.
We add a script to deploy generic tokens with parameters to specify "ckETH".

This will be used in a future PR to add ckETH canisters to a stock snapshot.

NOTE: Look at the second commit separately to see how the generic scripts differ from the ckBTC scripts.

# Changes

1. Make copies of ckBTC specific scripts.
2. Remove KYT, minter and other ckBTC specific code.
3. Add tests to CI.

### DIFF
--- a/bin/dfx-token-deploy
+++ b/bin/dfx-token-deploy
@@ -1,0 +1,133 @@
+#!/usr/bin/env bash
+# Original source: https://github.com/dfinity/ckBTC-Minter-Frontend/blob/master/local_deploy.sh
+# Via:             https://github.com/dfinity/nns-dapp/blob/main/scripts/ckbtc/deploy-ckbtc
+set -euo pipefail
+SOURCE_DIR="$(dirname "$(realpath "${BASH_SOURCE[0]}")")"
+PATH="$SOURCE_DIR:$PATH"
+. "$SOURCE_DIR/versions.bash"
+
+print_help() {
+  cat <<-EOF
+
+	Installs ckbtc canisters.
+
+	EOF
+}
+
+# Source the clap.bash file ---------------------------------------------------
+source "$SOURCE_DIR/clap.bash"
+# Define options
+clap.define short=n long=network desc="The dfx network to use" variable=DFX_NETWORK default="local"
+clap.define short=p long=prefix desc="Prefix for the local canister names" variable=LOCAL_PREFIX default="ckbtc_"
+clap.define short=y long=yes desc="Deploy even if there are existing ckbtc canisters." variable=DFX_YES nargs=0
+clap.define short=c long=check desc="Check that the canisters are present and correct." variable=DFX_CHECK nargs=0
+clap.define short=m long=mode desc="Canister install mode." variable=DFX_MODE default="reinstall"
+# Source the output file ----------------------------------------------------------
+source "$(clap.build)"
+
+: Check preconditions. This is always run.
+"${SOURCE_DIR}/dfx-ckbtc-import" --check --prefix "$LOCAL_PREFIX" || {
+  echo "ERROR: ckbtc canister data and files are not all present and correct."
+  echo "       Please run:"
+  echo "       dfx-ckbtc-import --prefix '$LOCAL_PREFIX'"
+  echo
+  exit 1
+} >&2
+
+deploy_ckbtc() {
+  # We create some canisters ahead of time because we need to know their IDs
+  # before deploying. Remaining canisters will be created when deployed.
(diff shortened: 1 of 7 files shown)


bin/dfx-token-deploy
```diff
@@ -0,0 +1,81 @@
...
+     token_name = \"Token $TOKEN\";
+     minting_account = record { owner = principal \"$DFX_IDENTITY_PRINCIPAL\" };
+     transfer_fee = 10;
+     max_memo_length = opt 64;
+     metadata = vec {};
+     initial_balances = vec { record { record { owner = principal \"$DFX_IDENTITY_PRINCIPAL\"; }; 10_000_000; }; };
+     archive_options = record {
+         num_blocks_to_archive = 10_000;
+         trigger_threshold = 20_000;
+         controller_id = principal \"$DFX_IDENTITY_PRINCIPAL\";
```
lmuntaner, line 60: Is this the same account as the ICP faucet? https://github.com/dfinity/nns-dapp/blob/d8568bd0c944a50e731fac29d8f638f4fc330234/frontend/src/lib/api/dev.api.ts#L31

bin/dfx-token-import
```diff
@@ -0,0 +1,125 @@
...
+      { "\(env.LOCAL_PREFIX)ledger": { build: "true", candid: "\(env.CANDID_DIR)/\(env.LOCAL_PREFIX)ledger.did", wasm: "\(env.WASM_DIR)/\(env.LOCAL_PREFIX)ledger.wasm", type: "custom" }
+      , "\(env.LOCAL_PREFIX)index" : { build: "true", candid: "\(env.CANDID_DIR)/\(env.LOCAL_PREFIX)index.did",  wasm: "\(env.WASM_DIR)/\(env.LOCAL_PREFIX)index.wasm",  type: "custom" }
+      }} * .[0]
+  ' dfx.json | sponge dfx.json
+
+  : Get the wasms
+  mkdir -p "$WASM_DIR"
+  get_wasm() {
+    local remote_name local_name local_path url
+    remote_name="$1"
```
lmuntaner, line 48: Stupid question, when is `remote_name` set?

lmuntaner (approved):
Thanks!

A couple of questions, but rubber stamping the bash code.

bin/dfx-token-import
dskloetd: It's set on this exact line that you commented on.
It's set with the value passed as the first parameter to `get_wasm`, for example on line 55 (`ic-icrc1-ledger.wasm`).

bin/dfx-token-deploy
dskloetd: No. This is the principal of the identity installing the canisters. So you can mint using a `dfx` command.
My plan was to then mint a lot of tokens into the faucet account (next PR).
It seems slightly more generic this way but if you're worried that the faucet will run dry, we can also make it the minting account.

lmuntaner: Ah, ok, no, it's ok. Whatever you think it's easier.

We can change later as well. We would only need to hardcode the faucet account principal here, right?

bin/dfx-token-import
```diff
@@ -0,0 +1,125 @@
...
+    local_name="$2"
+    local_path="$(c="$local_name" jq -re '.canisters[env.c].wasm' dfx.json)"
+    url="https://download.dfinity.systems/ic/$DFX_IC_COMMIT/canisters/${remote_name}.gz"
+    echo "Getting  $local_path from $url..."
+    curl -fsSL "$url" | gunzip >"${local_path}"
+  }
+  get_wasm ic-icrc1-ledger.wasm "${LOCAL_PREFIX}ledger"
```
lmuntaner, line 55: These are not ckETH specific, right?

Will there be specific ledger and index canisters or will they always use this one?

For example, there are specific ledger and index canisters for ckbtc.

bin/dfx-token-deploy
dskloetd: Yes, hardcoding the faucet address here would actually work because (and this surprised me) there is no separate method for minting. Minting is done by asking for a transfer from a minting account.

bin/dfx-token-import
dskloetd: These are not specific to ckETH nor ckBTC. They are just the standard ICRC-1 ledger and index.
ckBTC also does not use specific ledger/index canisters.
In fact, if you look at just the second commit in this PR, you see that this line you commented on is unchanged compared to the copied `dfx-ckbtc-deploy` script.

PR #284, opened 2023-11-28
Install ckETH canisters in stock deploy

dskloetd:
# Motivation

We want to support the ckETH token in the NNS dapp.
So we install an additional ICRC-1 token and call it ckETH to use in the test environment.

# Changes

1. Import and deploy a token and call it ckETH.
2. Mint 10M tokens into the faucet address such that "Get ckETH" on the NNS dapp will work.

# Tested

Manually: Created a snapshot and checked that I could get the new ledger IDs when running the snapshot.


lmuntaner (approved):
LGTM! Thanks!

PR #285, opened 2023-11-28
Make token name same as symbol

dskloetd:
# Motivation

In the universe selector in NNS dapp, ckETH is listed as "Token ckETH", which looks strange next to ckBTC (without "Token"). Apparently ckBTC doesn't get its token name from the ledger metadata but ckETH does.

# Change

Remove "Token" from the token name of the generic token deploy script and the ckbtc deploy script.


lmuntaner (approved):
Thanks!

The ckBTC on mainnet doesn't have the "Token" prefix.

https://dashboard.internetcomputer.org/canister/mxzaz-hqaaa-aaaar-qaada-cai

We should also use the name from there.

PR #290, opened 2023-12-07
Use index-ng for ckBTC

dskloetd:
# Motivation

There is a new index canister called index-ng.
For ckETH we had to use the new index canister because the old one does not support u256 balances.
For SNS the canisters are installed by SNS-W which we don't control.
But for ckBTC we were still using the old index canister.

# Changes

1. In `dfx-ckbtc-import` download the wasm and candid files for index-ng.
2. In `dfx-ckbtc-deploy` use the init arg format for index-ng.

# Tested

I created a snapshot and used it with nns-dapp.
I thought the reason that minter burn transactions don't have the spender field set was because we were still using the old index canister. But I tested it and the new index canister also doesn't seem to have the spender field set.
But it also doesn't seem to cause any problem and it seems good to use the new canister since mainnet ckBTC also uses it.

### DIFF
--- a/bin/dfx-ckbtc-deploy
+++ b/bin/dfx-ckbtc-deploy
@@ -104,7 +104,7 @@ deploy_ckbtc() {
 })" --mode="$DFX_MODE" ${DFX_YES:+--yes} --upgrade-unchanged
 
   echo "Step 4: deploying index canister..."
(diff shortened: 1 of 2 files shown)


lmuntaner (approved):
LGTM!

PR #295, opened 2023-12-11
Fix openssl_static_binaries download

dskloetd:
# Motivation

In snsdemo we download binaries with OpenSSL statically linked because they depend less on what is installed in the environment where they are run. These binaries used to be in a special download directory, but now all binaries are statically linked and the separate directory no longer exists. This means that our downloads were failing and we were not updating our IC COMMIT. It's currently 2 weeks old at `711369db94d9bd81d1f9127c54e644ab97a72fa3`.

# Changes

1. In `bin/dfx-software-ic-install-executable` download from the `binaries` directory instead of the `openssl-static-binaries` directory.
2. In `bin/dfx-software-ic-latest` change the files that we try to download:
    a. Check `binaries/` instead of `openssl-static-binaries/`
    b. Remove disk images. They also no longer exist at the location we try to download them from but also snsdemo doesn't use disk images so there is no point in checking if they exist.
    c. Add some canisters that snsdemo requires just to make sure they exist at the used commit. This didn't seem to be necessary before but I'm adding them in case the presence of the disk images made it more likely for other things to also exist at the same commit.

# Tested

1. Ran `bin/dfx-software-ic-latest -x ../../ic` and got `f45daff4e63ad2d02caa90d35a9144ba5ddddb25`
2. Installed the `ic-admin` from the `binaries` directory and checked that it worked on my Mac. Previously only the `openssl-static-binaries` version worked on my Mac.

### DIFF
--- a/bin/dfx-software-ic-install-executable
+++ b/bin/dfx-software-ic-install-executable
@@ -15,12 +15,12 @@ clap.define short=b long=bin desc="Diectory in which to install the executeble" 
 source "$(clap.build)"
 
 install_linux() {
-  URL="https://download.dfinity.systems/ic/${DFX_IC_COMMIT}/openssl-static-binaries/x86_64-linux/${EXEC_NAME}.gz"
+  URL="https://download.dfinity.systems/ic/${DFX_IC_COMMIT}/binaries/x86_64-linux/${EXEC_NAME}.gz"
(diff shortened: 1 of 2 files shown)


lmuntaner (approved):
Thanks!

PR #297, opened 2023-12-11
Use 1-proposal in snsdemo

dskloetd:
# Motivation

The 2-proposal flow is deprecated and the 1-proposal flow is simpler.

In snsdemo we have 2 ways of making SNSes:
1. Making a single SNS
2. Making multiple SNSes in parallel

This PR only changes the single SNS flow to 1-proposal. The parallel flow will be done in another PR.

The 2-proposal flow had a separate step to deploy SNS canisters, which made it easy to get their canister IDs. The 1-proposal flow does the deployment automatically so we also need a new script to import those canister IDs.

**Note**: The first commit copies some files that were used as a base for 2 new files to make it easier to see what has changed compared to the originals.

# Changes

1. Add a new `sns_init.yaml` file to use for the 1-proposal flow.
2. Add a flag to `bin/dfx-sns-config-random` to specify which SNS init template to use, defaults to the old 2-proposal file in order not to affect the parallel creation flow.
3. Add a new `bin/dfx-sns-propose` script based on the old `bin/dfx-sns-sale-propose`, with a flag to save the proposal ID to.
4. Add a new `bin/dfx-sns-import-by-proposal` script that imports the canister IDs created by the 1-proposal flow based on the proposal ID. It retries many times because the SNSes might not be created immediately.
5. Update `bin/dfx-sns-demo-mksns` to use the 1-proposal flow.

# Tested

1. Created a stock snapshot and ran it from nns-dapp.
2. Checked in the UI that 2 of the 12 SNSes were created with a 1-proposal
3. Checked that I was able to get tokens from the faucet for 1 of them.
4. Checked that I could participate in the swap of the other one.

### DIFF
--- a/bin/dfx-sns-demo-mksns
+++ b/bin/dfx-sns-demo-mksns
@@ -5,6 +5,13 @@ DEMO_BIN="$(realpath "${SOURCE_DIR}/../bin-other")"
 PATH="$DEMO_BIN:$SOURCE_DIR:$PATH:$(dfx cache show)"
 export PATH="$PATH:$HOME/.local/bin"
 
+print_help() {
+  cat <<-EOF
+
+	Creates a new random SNS proposal and imports the SNS canister IDs into dfx.json.
+	EOF
+}
+
 # Source the clap.bash file ---------------------------------------------------
 source "$SOURCE_DIR/clap.bash"
 # Define options
--- a/bin/dfx-sns-demo-mksns
+++ b/bin/dfx-sns-demo-mksns
@@ -15,17 +15,27 @@ source "$(clap.build)"
 set -x
 cd "$(dirname "$(realpath "$0")")/.."
 
-./bin/dfx-sns-whitelist-me --network "$DFX_NETWORK"
+: Generate the sns.yml file.
+./bin/dfx-sns-config-random --confirmation_text "${CONFIRMATION_TEXT:-}" --template_file "$SOURCE_DIR/sns_init.yaml"
 
-./bin/dfx-sns-config-random --confirmation_text "${CONFIRMATION_TEXT:-}"
+: Make sure we have enough resources.
 dfx ledger top-up --amount 2.0 --network "$DFX_NETWORK" "$(dfx identity get-wallet --network "$DFX_NETWORK")"
-./bin/dfx-sns-deploy --network "$DFX_NETWORK"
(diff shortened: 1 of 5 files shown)


bin/dfx-sns-propose
```diff
@@ -10,38 +10,11 @@ source "$SOURCE_DIR/clap.bash"
...
-  --target-swap-canister-id "$SNS_SWAP" \
-  --neuron-basket-count 2 \
-  --neuron-basket-dissolve-delay-interval-seconds 2629800 \
-  --proposal-title "$PROPOSAL_TITLE" \
-  --summary "Some summary"
-
-printf ' "%q"' "${@}"
-echo
-"${@}"
+bin-other/sns propose --neuron-id "$DFX_NEURON_ID" ${PROPOSAL_ID_FILE:+--save-to "$PROPOSAL_ID_FILE"} --network "$DFX_NNS_URL" sns.yml
```
lmuntaner, line 29: which step creates `sns.yml`? the `dfx-sns-config-random`?

bin/dfx-sns-demo-mksns
```diff
@@ -15,17 +22,27 @@ source "$(clap.build)"
...
+rm "$PROPOSAL_ID_FILE"
+
+: Import the SNS canister IDs for the created proposal.
+dfx-sns-import-by-proposal --proposal_id "$PROPOSAL_ID" --network "$DFX_NETWORK"
+
+: Output the imported canister IDs.
 dfx canister id sns_root --network "$DFX_NETWORK"
 dfx canister id sns_governance --network "$DFX_NETWORK"
 dfx canister id sns_ledger --network "$DFX_NETWORK"
+dfx canister id sns_index --network "$DFX_NETWORK"
```
lmuntaner, line 47: Does the script fail if any of these fail?

lmuntaner (approved):
Thanks!! LGTM. I have just a few questions.

bin/dfx-sns-demo-mksns
dskloetd: Yes, the `set -e` at the top makes the script fail if any of the commands in it fails.

bin/dfx-sns-propose
dskloetd: Correct.
On line 52:
```
sed "$SED_COMMANDS" "$SNS_INIT_PATH" >sns.yml
```
`>sns.yml` sends the output of the command to that file.

It's also documented in the help text of `dfx-sns-config-random`:
```
Creates an sns.yml to configure an SNS swap.
```

PR #298, opened 2023-12-12
Support 1-proposal in parallel SNS creation

dskloetd:
# Motivation

The 2-proposal flow is deprecated and the 1-proposal flow is simpler.

In snsdemo we have 2 ways of making SNSes:

1. Making a single SNS
2. Making multiple SNSes in parallel

[The previous PR](https://github.com/dfinity/snsdemo/pull/297) only changed the single SNS flow to 1-proposal.
This PR changes the parallel flow to use 1-proposal.

# Changes

1. In `bin/dfx-sns-config-random` use the new `sns_init.yaml` template by default.
2. In `bin/dfx-sns-demo-mksns` stop passing the `--template_file` argument now that the default is to use the new template.
3. Update `bin/dfx-sns-demo-mksns-parallel` to follow the same process as `bin/dfx-sns-demo-mksns`: First open all the swaps in parallal. Then close all the swaps in parallel. It's now possible to do both parts in parallel, making parallel SNS creation even faster.
4. Update the README to refer to the new `sns_init.yaml`.
5. Delete files which were specific to the 2-proposal flow.

Small unrelated fixes that made development easier:

1. Set `PATH` in `bin/dfx-neuron-vote` (similar to other scripts in snsdemo) so it doesn't fail to call `dfx-neuron-id` if the user doesn't have their `PATH` set.
2. In `bin/dfx-sns-aggregator-wait` don't output an error if one of the aggregator pages doesn't exist yet.

# Tested

Created a snapshot and ran it in nns-dapp. Participated in a swap and took tokens out of the faucet. Checked there were 12 SNSes.


lmuntaner (approved):
Thanks!!

PR #308, opened 2024-01-09
Update dfx to 0.15.3

dskloetd:
# Motivation

We want to keep the dfx version of our test environment up to date.
Additionally the newest SNS governance canister requires a new version of dfx because it's using a new type of performance counter.
We haven't been able to use any version after 0.14.4 because 0.15 didn't support connecting remotely, which we use for testing with DevEnv.
This has now been fixed with version 0.15.3

Additionally since version 0.15 the `dfx nns` and `dfx sns` commands are no longer supported by default and require installing respective extensions.

# Changes

1. Change dfx version from 0.14.4 to 0.15.3.
2. Add scripts to install `nns` and `sns` extensions.
3. Call the scripts in `dfx-sns-demo-install`.

# Tested

1. Created a snapshot on DevEnv and connected to it from my laptop.
2. Ran `dfx-sns-demo --ic_commit latest --x ../../ic` to make sure that the SNS governance canister can be used again.


lmuntaner (approved):
LGTM!

PR #310, opened 2024-01-10
Don't fall back on pinned version when '--ic_commit latest' is specified

dskloetd:
# Motivation

When running the demo or creating a snapshot, we use a pinned IC commit for all the canisters.
But it's also possible to specify `--ic_commit latest` to use the latest version.
When you do, you also have to specify a directory that holds the IC repo so that it can find the latest commit.
If you specify a wrong `--ic_dir` or the default value is not valid for you, currently it falls back to using the pinned version which might be very confusing.

# Changes

When trying to find the latest IC commit, if the specified directory doesn't exist or is not a git repository, explicitly fail instead of returning nothing, which later results in using the pinned version.

# Tested

I tried
```
$ bin/dfx-sns-demo --ic_commit latest
```
and
```
$ bin/dfx-snapshot-stock-make --ic_commit latest
```
both with and without my changes.
Without my changes, they fall back on the pinned version and with my changes they fail.
If I additionally specify `--ic_dir ../../ic`, they work as intended.


lmuntaner (approved):
LGTM!

PR #322, opened 2024-02-19
Disable automatic update of dfx version

dskloetd:
# Motivation

Having to update dfx very frequently is somewhat painful.
Especially since snapshots don't work between versions.
Also there is limited benefit to always being on the latest version unless we need a specific feature.

# Changes

Remove the scheduled running of the dfx update job.
We keep the job so it can be run manually if we want.


lmuntaner (approved):
Thanks

PR #324, opened 2024-02-21
SNSDEMO: Fix CI, use GitHub action to install dfx

dskloetd:
# Motivation

The dfx install script at https://sdk.dfinity.org/install.sh changed and now installs dfxvm (version manager) instead of dfx itself and also requires interaction. This broke the installation of dfx on our CI jobs.

There is a GitHub actions provide which also installs dfx.

# Changes

In our CI jobs that requires dfx installed, use the GitHub action before `bin/dfx-sns-demo-install` gets called, such that we are not forced into the interactive install flow.

### DIFF
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -75,7 +75,7 @@ jobs:
         env:
           GH_TOKEN: ${{ github.token }}
       - name: Install dfx
-        run: ./bin/dfx-software-dfx-install --version "$(jq -r .dfx dfx.json)"
+        uses: dfinity/setup-dfx@main
       - name: "Test the aggregator wasm install command"
         run: ./bin/dfx-software-sns-aggregator-install.test --verbose
         env:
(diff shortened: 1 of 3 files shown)


lmuntaner (approved):
LGTM

bitdivine (approved):

.github/workflows/checks.yml
```diff
@@ -75,7 +75,7 @@ jobs:
         env:
           GH_TOKEN: ${{ github.token }}
       - name: Install dfx
-        run: ./bin/dfx-software-dfx-install --version "$(jq -r .dfx dfx.json)"
+        uses: dfinity/setup-dfx@main
```
bitdivine, line 78: Just checking - this does get the version from dfx.json, yes?

bitdivine: Yes, it is: https://github.com/dfinity/setup-dfx

dskloetd: Yes, since yesterday, on my request.

PR #325, opened 2024-02-22
Fix snsdemo snapshot for dfxvm

dskloetd:
# Motivation

When using `dfx` with `dfxvm`, there some extra required configuration to determine which version of `dfx` to use.
This affects snapshots in 2 ways:

1. When you start a snapshot, this configuration is replaced by the snapshot, so if the snapshot does not contain it, you will get an error when running dfx because it doesn't know which version to use.
2. The script to create SNSes in parallel creates separate users and these users need this configuration to use `dfx`. This already worked on Linux because a symbolic link of the config directory was created. But the config directory is different on Mac.

# Changes

1. Include `version-manager.json` in the snapshot, so that when the snapshot is used, `dfx` knows which version to use.
2. Use the correct directory on Mac to create a symbolic link for the config.

# Tested

Manually tested on DevEnv and Mac. Tested on CI.

### DIFF
--- a/bin/dfx-snapshot-save
+++ b/bin/dfx-snapshot-save
@@ -45,7 +45,10 @@ IFS=', ' read -r -a DFX_IDENTITIES <<<"$DFX_IDENTITIES"
 
 # Get global state
 pushd "$HOME"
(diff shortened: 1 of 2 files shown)


mstrasinskis (approved):
Thank you!

PR #342, opened 2024-05-08
Generate human readable sns names

mstrasinskis:
# Motivation

We would like to have Snses with human readable names to make our presentation look nicer. 

# Changes

- Script to generate a token name (1 or 2 words).
- Script to generate a token symbol from token name.
- Utilise new scripts in `dfx-sns-config-random`.

# Screenshots

Screenshot 2024-05-08 at 16 13 53


bin/dfx-sns-config-random-token-symbol
```diff
@@ -0,0 +1,11 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+NAME="$1"    # Full name passed as the first argument
+COMPACTED="" # Initialize the compacted string without spaces
+
+# Remove spaces and convert to uppercase
+COMPACTED=$(echo "$NAME" | tr -d ' ' | tr '[:lower:]' '[:upper:]')
+
+# Output the first 3 to 5 characters (adjust as needed)
```
dskloetd, line 10: It just outputs the first 5, right? Why does it say 3?

bin/dfx-sns-config-random-token-symbol
```diff
@@ -0,0 +1,11 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+NAME="$1"    # Full name passed as the first argument
+COMPACTED="" # Initialize the compacted string without spaces
```
dskloetd, line 5: This line is not necessary. Just initialize the value directly on line 8.

bin/dfx-sns-config-random-token-symbol
dskloetd, line 1: This doesn't do anything random.
I'm not sure this even warrants a separate script, but if so, it should have some usage information.
But easiest to just inline it as 2 lines where it's used.

mstrasinskis:
Replaced with https://github.com/dfinity/snsdemo/pull/370

PR #343, opened 2024-05-08
More sns logos

mstrasinskis:
# Motivation

We would like to have demo Snses to look nicer.

# Changes

- Added a bunch of images (some are free, some are from our figma).
- Chose one of the available image for the sns logo.
- Use different images for logo and token.

# Screenshots

image

image


demo-assets/20.png
dskloetd, line 1: How come this looks like a low resolution PNG scaled up, stored as JPEG and then converted back to PNG?

bin/dfx-sns-config-random-token-logo
```diff
@@ -0,0 +1,41 @@
+#!/usr/bin/env bash
+set -euo pipefail
+SOURCE_DIR="$(dirname "$(realpath "${BASH_SOURCE[0]}")")"
+REPO_ROOT=$(git -C "$(dirname "$SOURCE_DIR")" rev-parse --show-toplevel)
+
+# Define the directory containing the images
+DIR=$(realpath "$REPO_ROOT/demo-assets")
+
+# File to store the current index
+INDEX_FILE="$DIR/_logo_current_index.txt"
```
dskloetd, line 10: I don't see this file anywhere.

bin/dfx-sns-config-random-token-logo
dskloetd, line 1: I don't see anything random. What is this file really trying to do?

demo-assets/1.png
dskloetd, line 1: We should probably keep a record of where you got each of these images so we're sure we can use them, right?

mstrasinskis:
Replaced with https://github.com/dfinity/snsdemo/pull/370

PR #346, opened 2024-05-10
Make it easier to create a manual snsdemo release

dskloetd:
# Motivation

snsdemo releases are automatically created once a week.
In order to create a release outside the schedule, one has to perform some manual steps.
Since we already have a workflow that can create a release let's also use that for manual release.

# Changes

1. Add a checkbox when starting the workflow manually to indicate a release should be created.
2. Create a release when the checkbox is checked in addition to when the workflow runs on schedule.
3. Remove the functionality to create a release from branch `snapshot-schedule`. This was a branch created by @bitdivine supposedly for testing.
4. If a release was already created on the same day, try appending `.1` or `.2` etc., similar to what ic-js and gix-components do.

# Tested

Checkbox appears:
image

Created this release from this branch: https://github.com/dfinity/snsdemo/releases/tag/release-2024-05-10

### DIFF
--- a/.github/workflows/snapshot.yml
+++ b/.github/workflows/snapshot.yml
@@ -48,7 +48,7 @@ jobs:
       - name: Stop dfx
         run: dfx stop
       - name: Tag if release should be created
-        if: ${{ (github.event_name == 'schedule') || (github.inputs.make_release == true)}}
+        if: ${{ (github.event_name == 'schedule') || (inputs.make_release)}}
         id: tag
         run: |
           set -euxo pipefail
--- a/.github/workflows/snapshot.yml
+++ b/.github/workflows/snapshot.yml
@@ -2,6 +2,11 @@ name: Create a snapshot image
 on:
   push:
   workflow_dispatch:
+    inputs:
+      make_release:
+        description: 'Make release'
+        default: false
+        type: boolean
   schedule:
     - cron: '30 3 * * WED'
 env:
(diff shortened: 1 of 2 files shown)


.github/workflows/snapshot.yml
```diff
@@ -42,20 +47,22 @@ jobs:
...
-            git tag "${tag_name}"
-            git push origin "refs/tags/${tag_name}"
-          fi
+          index=0
+          while git show-ref --tags --verify --quiet "refs/tags/${tag_name}"; do
+            echo "Tag ${tag_name} already exists."
+            index=$((index + 1))
+            tag_name="$(date +release-%Y-%m-%d).${index}"
+            echo "Trying tag ${tag_name}."
+          done
```
daniel-wong-dfinity-org, line 62: If possible, comment the purpose/effect of this loop.

```suggestion
          # If there is already a tag with the same name, try appending ".${next_integer}".
          while git show-ref --tags --verify --quiet "refs/tags/${tag_name}"; do
            echo "Tag ${tag_name} already exists."
            index=$((index + 1))
            tag_name="$(date +release-%Y-%m-%d).${index}"
            echo "Trying tag ${tag_name}."
          done
```

PS: This seems orthogonal to the title. Maybe mention this too?

README.md
```diff
@@ -37,14 +37,9 @@ the parameters passed to `ic-admin` from `bin/dfx-sns-sale-propose`.
...
 
-A new stock snapshot can be released by pushing a tag:
-
-```
-RELEASE="release-$(date +"%Y-%m-%d")"
-echo "$RELEASE"
-git tag "$RELEASE"
-git push origin "$RELEASE"
-```
+A new stock snapshot can be released by running the
```
daniel-wong-dfinity-org, line 40: No change.

I know the previous version said this, but I'm not sure what "stock snapshot" means.

Did the old instructions not work?

daniel-wong-dfinity-org (approved):

.github/workflows/snapshot.yml
dskloetd: Do you mean the PR title?
It's mentioned in the PR description, but it's not orhogonal.
If you only have scheduled releases on different dates, the tag names based on dates never collide.
But if you can create releases any time, you might create multiple on the same day, so we need this collision avoidance.

I've applied your comment suggestion. Thanks.

README.md
dskloetd: "stock snapshot" is what we call the snsdemo snapshot in some places.
I guess the thought was that we'd have custom snapshots as well and have one "stock" snapshot which isn't custom.

The old instructions still work but the new instructions are simpler so it didn't seem useful to keep the old instructions.

PR #350, opened 2024-05-16
Update nns and sns extensions pinned version

dskloetd:
# Motivation

We haven't updated the version of the `nns` and `sns` extensions we use in some time.
It's not possible to upgrade an extension, so we used to be stuck with one version but now it's at least possible to uninstall and reinstall.

# Changes

1. Add `bin/dfx-software-dfx-extension-install` to hold the shared code between `bin/dfx-software-dfx-extension-nns-install` and `bin/dfx-software-dfx-extension-sns-install`.
2. Check the current installed version and uninstall if it's different from the desired version.

Note that when the "latest" version is requested, we don't ask for a specific version but take what we get. We don't know ahead of time what we'll get so in this case we always uninstall first.

# Tested

Tested manually by installing with `--version 0.2.1`, with `--version latest` and `--version pinned`.


bin/dfx-software-dfx-extension-install
```diff
@@ -0,0 +1,51 @@
...
+  if [[ "$current_version" = "$VERSION" ]]; then
+    echo "$EXTENSION extension version $VERSION is already installed."
+    exit 0
+  fi
+  echo "Uninstalling current $EXTENSION extension version $current_version."
+  dfx extension uninstall "$EXTENSION"
+fi
+
+dfx extension install "$EXTENSION" "${VERSION_ARGS[@]}"
+echo "Installed $EXTENSION extension version $(get_current_version)"
```
mstrasinskis, line 51: Since this is the only output when there are no errors, how about mentioning dfx?
```suggestion
echo "Installed $EXTENSION dfx extension version $(get_current_version)"
```

mstrasinskis (approved):

PR #351, opened 2024-05-16
Add ckUSDC canisters to snapshot

dskloetd:
# Motivation

We want to add ckUSDC to NNS dapp and should have a way to test it.

# Changes

1. Add a logo file (copied from metadata of current mainnet ckSepoliaUSDC).
2. Add a flag to `bin/dfx-token-deploy` to allow specifying a logo file.
3. Create ckUSDC canisters in `bin/dfx-stock-deploy`.
4. Add 10M ckUSDC to the faucet.

# Tested

Created a snapshot and used it in NNS dapp


bin/dfx-token-deploy
```diff
@@ -47,6 +48,12 @@ deploy_token() {
 
   echo "Step 1: deploying ledger canister..."
   DFX_IDENTITY_PRINCIPAL="$(dfx identity get-principal)"
+  if [[ -n "${LOGO_SVG_FILE:-}" ]]; then
+    LOGO_BASE64="$(base64 -i "$LOGO_SVG_FILE")"
+    TOKEN_METADATA="vec { record { \"icrc1:logo\"; variant { Text = \"data:image/svg+xml;base64,${LOGO_BASE64}\" }; } }"
+  else
+    TOKEN_METADATA="vec {}"
```
mstrasinskis, line 55: Will it work w/o a logo?

mstrasinskis (approved):

bin/dfx-token-deploy
dskloetd: Yes if you look at line 58 on the previous version of the file, that's how it used to be and that's how ckETH is installed.
But ckETH has a hard-coded logo in nns-dapp so you don't notice that the canister doesn't have a logo.
If you neither have a hardcoded nor a logo in the canister, NNS dapp shows a question mark logo.

PR #376, opened 2024-07-08
Add ckRED canisters to snapshot

mstrasinskis:
# Motivation

For the imported token feature we need a custom token for testing.

# Changes

1. Add a logo file (generated new svg).
3. Create ckRED canisters in `bin/dfx-stock-deploy`.
4. Add 10M ckRED to the faucet.

# Tested

Created a snapshot and used it in NNS dapp

image

### DIFF
--- a/bin/dfx-stock-deploy
+++ b/bin/dfx-stock-deploy
@@ -76,7 +76,7 @@ dfx-token-import --prefix ckred_ --commit "$DFX_IC_COMMIT"
 dfx-token-deploy \
   --prefix ckred_ \
   --token ckRED \
-  --decimals 6 \
+  --decimals 7 \
   --fee 4000 \
   --logo-svg-file "$SOURCE_DIR/../ckRED-logo.svg" \
   --network "$DFX_NETWORK" \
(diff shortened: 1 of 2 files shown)


ckRED-logo.svg
dskloetd, line 1: Nice logo!

bin/dfx-stock-deploy
```diff
@@ -71,13 +71,25 @@ dfx-token-deploy \
   --network "$DFX_NETWORK" \
   --yes
 
+: Set up ckRED canisters
+dfx-token-import --prefix ckred_ --commit "$DFX_IC_COMMIT"
+dfx-token-deploy \
+  --prefix ckred_ \
+  --token ckRED \
+  --decimals 6 \
```
dskloetd, line 79: Do we want 6 decimals? I guess this is copied from ckUSDC.
Standard is 8. If we want non-standard because it gives better test coverage, should we pick something other than 6?

I don't have a strong opinion, just wondering if you thought about it at all.

dskloetd (approved):

bin/dfx-stock-deploy
mstrasinskis: Yep, I thought about it, and I think anything with a non-default value will work. But I’m not against changes; let’s use 7.

PR #381, opened 2024-07-22
test PR

dskloetd:
Testing https://github.com/dfinity/snsdemo/pull/380


yhabib:
This was a test for #380, which has already landed.

PR #382, opened 2024-07-23
Get number of SNSes from SNS-WASM canister

dskloetd:
# Motivation

When creating a snapshot, before killing the replica and archiving the snapshot, we wait for the aggregator to scrap all SNSes.
Currently we specify to the aggregator how many SNSes it should expect.
This is not necessary because we can get this information from the SNS-WASM canister.
And it can lead to unnecessary delays when the number is wrong.

# Changes

Instead of passing the number of expected SNSes to the `dfx-sns-aggregator-wait` let the script determine the expected number for itself by calling the SNS-WASM canister.

# Tested

CI passes and correctly detects 12 SNSes:
https://github.com/dfinity/snsdemo/actions/runs/10057983124/job/27800205191?pr=382
image


lmuntaner (approved):
Nice!

PR #378, opened 2024-07-19
Update dfx to version 0.21.0

sa-github-api:
Automated changes by [create-pull-request](https://github.com/peter-evans/create-pull-request) GitHub action

PR #383, opened 2024-07-23
Bump dfx to 0.21.0

dskloetd:
# Motivation

The automatic dfx version increase [PR](https://github.com/dfinity/snsdemo/pull/378) failed because the code we use to check if a checkpoint was created failed.

Version 0.21.0 of dfx can now have multiple state directories for different versions and the config file that tells us where to find the state directory is no longer there.

# Changes

1. Instead of trying to find the specific state directory to look for checkpoints, look for checkpoints in all state directories. We assume only a single replica is running so only a single state directory will be receiving new checkpoints.
2. Update the dfx version in `dfx.json`.

# Tested

CI passes: https://github.com/dfinity/snsdemo/actions/runs/10058043110/job/27800272383?pr=383


lmuntaner (approved):
LGTM

PR #390, opened 2024-08-19
Update IC commit and extension versions

dskloetd:
# Motivation

We have an automated job to update the IC commit in snsdemo.
The latest PR had test failures:
https://github.com/dfinity/snsdemo/pull/389
https://github.com/dfinity/snsdemo/actions/runs/10446677877/job/28924422913?pr=389

The failure were caused by wasms being too large to install.
I tracked the cause down to https://github.com/dfinity/ic/commit/1e5a4012d2f0e7c649990db78da8eba42e575bc9
which suggests the solution is to use a newer version of the `nns` extension.

Additionally, `nns` extension versions from 0.4.1 fail if run before the replica is ready with:
```
Error: An error happened during communication with the replica: error sending request for url (http://127.0.0.1:8080/api/v2/status)
Error: Extension exited with non-zero status code '1'.
```
This can be worked around by using `dfx ping --wait-healthy` before calling the extension.

# Changes

1. Update the IC commit to `b0aef30f10784189a5ce73e58ed1f19359a5a3ae`, which is the same as the bot PR.
2. Update the extension versions from 0.4.0 to 0.4.3.

# Tests

CI passes again.


mstrasinskis (approved):
Thank you!

PR #401, opened 2024-09-16
Install SNS wasms without gunzipping

dskloetd:
# Motivation

https://github.com/dfinity/ic/commit/7f0f5d5d3f2d7a827158daf8b20530c81e3fd46c caused the uncompressed SNS governance wasm to be too large to be uploaded to SNS-W.
This results in ([logs](https://github.com/dfinity/snsdemo/actions/runs/10877138238/job/30178056736?pr=400)):
```
Caused by: The replica returned an HTTP Error: Http Error: status 413 Payload Too Large, content type "text/plain; charset=utf-8", content: Request 0x873d0da0b95934042af97fe6d220608b3ce8aeb8aca343ef5f06a53c404d03c8 is too large. Message byte size 3687223 is larger than the max allowed 3670016.
```
when [updating the IC commit](https://github.com/dfinity/snsdemo/pull/400).

This happens because we decompress wasms before uploading them but there is no need to do that.

# Changes

1. Do not `gunzip` downloaded SNS wasms but upload the compressed wasms directly.
2. Update the wasm filenames in `bin/sns_dfx.json` to be what we actually store.

# Tested

I tried running `bin/dfx-sns-demo` with these changes with IC commit `d4ee25b0865e89d3eaac13a60f0016d5e3296b31` which caused the failures in https://github.com/dfinity/snsdemo/pull/400 and it no longer failed.

### DIFF
--- a/bin/dfx-sns-wasm-download
+++ b/bin/dfx-sns-wasm-download
@@ -16,5 +16,5 @@ export DFX_IC_COMMIT
 
 if [[ "${1:-wasms}" == "wasms" ]]; then
   mkdir -p wasms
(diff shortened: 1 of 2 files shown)


mstrasinskis (approved):
Thank you!

PR #409, opened 2024-10-11
Wait for SNS lifecycle

dskloetd:
# Motivation

SNS swap used to update its lifecycle every second. Now it only does so once a minute.
When buying into the swap, or when finalizing the swap, it needs to be in the correct lifecycle.

# Changes

1. Add a script to get the current lifecycle of the current SNS.
2. Add a script to wait for the current SNS to be at a given lifecycle.
3. Use the script to make sure we are at the correct lifecycle before buying or finalizing.

# Tested

Tried creating a snapshot with the IC commit which has the new swap canister. It failed without these changes and passes with these (and other, upcoming) changes.

### DIFF
--- a/bin/dfx-sns-wait-for-lifecycle
+++ b/bin/dfx-sns-wait-for-lifecycle
@@ -4,6 +4,8 @@ SOURCE_DIR="$(dirname "$(realpath "${BASH_SOURCE[0]}")")"
 PATH="$SOURCE_DIR:$PATH"
 
 print_help() {
+  # ShellCheck somehow thinks this code is unreachable.
+  # shellcheck disable=SC2317
   cat <<-EOF
 
 	Wait up to 2 minutes for the current SNS to reach the given lifecycle.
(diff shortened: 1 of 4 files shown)


bin/dfx-sns-wait-for-lifecycle
```diff
@@ -0,0 +1,59 @@
...
+  "adopted"
+)
+
+LIFECYCLE_LIST=$(printf " %s," "${LIFECYCLES[@]}")
+LIFECYCLE_LIST=${LIFECYCLE_LIST%,*}
+
+# Source the clap.bash file ---------------------------------------------------
+source "$SOURCE_DIR/clap.bash"
+# Define options
+clap.define short=l long=lifecycle desc="One of $LIFECYCLE_LIST" variable=LIFECYCLE default="open"
```
mstrasinskis, line 29: Not sure this script needs a default lifecycle value, but up to you.

mstrasinskis (approved):
Thanks!

bin/dfx-sns-wait-for-lifecycle
dskloetd: Otherwise it needs extra logic to handle an empty lifecycle.
Seems harmless like this.

PR #410, opened 2024-10-11
Update IC repo candid paths

dskloetd:
# Motivation

Ledger and Index Candid files were moved to a different directory in https://github.com/dfinity/ic/pull/1682

# Changes

1. Update the IC commit to one where the candid files have been moved.
2. Download the candid files from the new paths.

# Tested

CI passes


mstrasinskis (approved):
Thank you!

PR #412, opened 2024-10-14
Install nns-dapp in dfx-stock-deploy

dskloetd:
# Motivation

`dfx nns install` installs an old version of nns-dapp. For the longest time this has been good enough to use as a backend for local development. The only thing it didn't support (AFAIK) was renaming canisters.

But recent features import token and TVL require more recent functionality from the backend so now an snsdemo snapshot can't be used for nns-dapp frontend development without first installing the nns-dapp canisters.

**NOTE:** bin/dfx-software-nns-dapp-install is modeled after `bin/dfx-software-sns-aggregator-install` from which it is copied in the first commit. So if you skip the first commit when reviewing, then you only need to review the differences between them.

# Changes

1. Set canister ID of `nns-icp-index` in `bin/dfx-network-deploy` because it isn't set by `dfx nns import` and is needed to pass in the argument of `nns-dapp`.
2. Update the pinned release of nns-dapp to the newest proposal.
3. Add `bin/dfx-software-nns-dapp-install` to install nns-dapp.
4. Call `bin/dfx-software-nns-dapp-install` in `bin/dfx-stock-deploy` after everything else is installed.

# Tested

Manually created a snapshot and used it with nns-dapp. TVL and import token worked without upgrading nns-dapp.


mstrasinskis (approved):
Looks good, thank you!

PR #419, opened 2024-11-04
Create wallet canisters sequentially

dskloetd:
# Motivation

Since `dfx` 0.24.1, creating the snapshot results in creating only 11 or 10 SNSes instead of 12 almost every time.
For example: https://github.com/dfinity/snsdemo/actions/runs/11659664378/job/32460602932?pr=418

After debugging, I found out that creating wallet canisters as part of the parallel SNS creation started occasionally failing with.
```
Error: Failed to setup wallet caller.
Caused by: Failed to create wallet
Caused by: Failed to load wallet configuration
Caused by: failed to parse contents of /home/runner/work/snsdemo/snsdemo/users/snsdemo_1/.local/share/dfx/network/local/wallets.json as json
Caused by: EOF while parsing a value at line 1 column 0
```

I'm not sure why but it might not work well anymore to create many wallet canisters simultaneously.

Also, while I was debugging this it appears that MacOS-12 stopped being supported on GitHub.

# Changes

1. Move the creating the snsdemo identities (which includes wallet canister creation) outside the parallel process so that it is done sequentially.
2. Check that we have the expected number of SNSes after SNS creation is done.
3. Use MacOS 14 instead of MacOS 12.

### DIFF
--- a/.github/workflows/run.yml
+++ b/.github/workflows/run.yml
@@ -12,7 +12,7 @@ jobs:
   #    strategy:
   #      fail-fast: false
   #      matrix:
-  #        os: [ macos-12, ubuntu-22.04 ]
+  #        os: [ macos-14, ubuntu-22.04 ]
   #    steps:
   #      - name: Checkout
   #        uses: actions/checkout@v3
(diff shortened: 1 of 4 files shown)


mstrasinskis (approved):
Thank you!

PR #420, opened 2024-11-06
Enforces an exact match of replica processes

yhabib:
A script runs `pgrep -l replica` to check for replica processes, but it can't receive false positives from a process called `replicatord`.

Eg:

```sh
$ pgrep -l replica
781 replicatord
95113 replica
```

This change allowed me to run it without issues. I also tried to start a second snapshot, and it correctly informed me that a replica was already running.

Related to https://github.com/dfinity/nns-dapp/pull/5732


dskloetd (approved):
Thanks!

PR #424, opened 2024-11-18
fix: Fix ckbtc minter wasm commit version to avoid installation problem

ninegua:
The ckBTC minter is under-going an upgrade that replaces the previous KYT canister with a new one.
Before the upgrade completes, we will pin the ckBTC installation to the last known commit that works.

### DIFF
--- a/bin/dfx-ckbtc-import
+++ b/bin/dfx-ckbtc-import
@@ -23,6 +23,9 @@ clap.define long=candid_dir desc="Where to store the imported candid files" vari
 # Source the output file ----------------------------------------------------------
 source "$(clap.build)"
 
+# NOTE: Fix the commit for ckbtc wasms to avoid installation problem. Remove this when it is no longer needed.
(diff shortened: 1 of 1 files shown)


bin/dfx-ckbtc-import
```diff
@@ -23,6 +23,9 @@ clap.define long=candid_dir desc="Where to store the imported candid files" vari
 # Source the output file ----------------------------------------------------------
 source "$(clap.build)"
 
+# NOTE: Fix the commit for ckbtc wasms to avoid installation problem. Remove this when it is no longer needed.
+DFX_IC_COMMIT=bcbccf79c89a3e81b1a38d8233f8f81d1af1b245
```
dskloetd, line 27: We have the `DFX_IC_COMMIT` pinned here:
https://github.com/dfinity/snsdemo/blob/f7b266c557d20a2de99e7184ae06fb893cb06352/bin/versions.bash#L5

We can rely on that for a short while.

PR #425, opened 2024-11-20
chore(IDX): replace PAT with GitHub Apps

cgundy:
This is part of a larger effort to replace PATs generated my service accounts with GitHub Apps.

https://github.com/dfinity/ic/pull/2717

### DIFF
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -18,7 +18,15 @@ jobs:
       - name: Check if commits can be added
         id: check_can_add_commit
         run: |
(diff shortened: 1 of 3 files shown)


dskloetd:
Do you have more information about this larger effort that you can link to?

dskloetd:
I was kind of assuming there was some document where the pros and cons were listed and a decision was documented. Even if it's not externally accessible, it's still useful to have for reference.

dskloetd (approved):

dskloetd:
Why don't you want to link it here? I think it would be good to link it here.

PR #427, opened 2024-11-25
chore: switch to ic-btc-kyt canister

ninegua:
The previous ic-ckbtc-kyt canister is now deprecated. A new ic-btc-kyt canister will be used instead.


dskloetd (approved):
Thanks!

PR #429, opened 2024-11-25
Use PocketIC in snsdemo

dskloetd:
# Motivation

Using PocketIC will eventually be the default, and then the only option, with `dfx start`.
But using it sooner will help with being able to control SNS canisters and making them run out of cycles in the test environment.

# Changes

1. Pass `--pocketic` when starting dfx to create a snapshot.
2. Using `--pocketic` when starting dfx to run a snapshot.
3. Stop waiting for a checkpoint. PocketIC will write a checkpoint when it's being stopped. And waiting for one earlier will just time out.

# Tests

Tested manually starting the snsdemo snapshot from nns-dapp on Mac and Linux.

### DIFF
--- a/bin/dfx-network-deploy
+++ b/bin/dfx-network-deploy
@@ -54,7 +54,7 @@ if [[ "$DFX_NETWORK" == "local" ]]; then
     tput setaf 0
   )"
   export NORMAL="$(tput sgr0)"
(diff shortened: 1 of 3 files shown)


yhabib (approved):
awesome

PR #430, opened 2024-11-28
Provide fake ICP Swap data in test environment

dskloetd:
# Motivation

We want to start providing total portfolio value in NNS dapp.
For this we want to use exchange rate date from ICP Swap as provided at https://uvevg-iyaaa-aaaak-ac27q-cai.raw.ic0.app/tickers
So for local testing we want to have similar data available for some locally installed tokens, in the same format.

# Changes

1. Add `bin/dfx-mock-icp-swap-install` which:
    1. Creates a JSON file with ticker data for installed ck and SNS tokens
    2. Adds an entry for an asset canister name `icp-swap` in `dfx.json`
    3. Uses `dfx deploy icp-swap` to deploy the asset canister with the tickers JSON file.
2. Move waiting for the aggregator to see all the SNSes from `bin/dfx-snapshot-stock-make` to `bin/dfx-stock-deploy` because `bin/dfx-mock-icp-swap-install` relies on the SNS aggregator to be up-to-date
3. Call `bin/dfx-mock-icp-swap-install` from `bin/dfx-stock-deploy`.

# Tests

1. Added a simple sanity check in `bin/dfx-stock-test`.
2. Tested manually by creating a snapshot and checking the URL as well as fetching the URL from NNS dapp with local changes.

### DIFF
--- a/bin/dfx-mock-icp-swap-install
+++ b/bin/dfx-mock-icp-swap-install
@@ -1,0 +1,94 @@
+#!/usr/bin/env bash
+set -euo pipefail
+SOURCE_DIR="$(dirname "$(realpath "${BASH_SOURCE[0]}")")"
+PATH="$SOURCE_DIR:$PATH"
+
+print_help() {
+  cat <<-EOF
+
+	Installs an asset canister which serves a JSON file at /tickers with fake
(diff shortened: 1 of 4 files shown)


yhabib (approved):
Thanks

PR #435, opened 2024-12-10
Rename btc kyt canister

dskloetd:
# Motivation

The IC commit wasn't updating past `9f5674ce6ba13379b24d2e49663d6dae087ae247` because it couldn't find `canisters/ic-btc-kyt.wasm.gz` because it was renamed in https://github.com/dfinity/ic/pull/2966

# Changes

1. Update IC commit
2. Rename kyt to checker.

# Tests

CI passes


mstrasinskis (approved):
Thanks a lot! Do we need to keep the other `kyt` occurrences?

dskloetd:
> Do we need to keep the other `kyt` occurrences?

We don't need to. We could change them to `checker` as well if we want.

PR #438, opened 2025-01-03
Use static fake ICP Swap prices instead of random

dskloetd:
# Motivation

In screenshot test in NNS dapp we need to mask out the dollar values because they differ per snapshot.
This makes testing tedious and reduces covergage.

# Changes

1. Provide static token prices in the fake ICP Swap response.

# Tests

Tested manually. It results in 1 ICP being $9.00, 1 ckBTC being $90'000, 1 ckETH being $3'000, 1 ckUSDC being $1.00 and other tokens being $0.18 each.
While this is somewhat realistic, it's actually not important that these prices are realistic. Is just nice to have some variety in prices for testing.

### DIFF
--- a/bin/dfx-mock-icp-swap-install
+++ b/bin/dfx-mock-icp-swap-install
@@ -27,11 +27,25 @@ if ! [[ -e dfx.json ]]; then
   exit 1
 fi
 
-get_random_price() {
-  # $RANDOM generates a random number between 0 and 32767.
-  # So we end up with a random number between 1 and ~17, with 4 digits behind
-  # the floating point.
-  echo "scale=4; $RANDOM / 2000 + 1" | bc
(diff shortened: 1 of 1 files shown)


yhabib (approved):
Thanks

bin/dfx-mock-icp-swap-install
```diff
@@ -27,11 +27,25 @@ if ! [[ -e dfx.json ]]; then
...
+    echo "9.00"
+    ;;
+  ckbtc_ledger)
+    echo "0.0001"
+    ;;
+  cketh_ledger)
+    echo "0.003"
+    ;;
+  *)
+    echo "50"
```
yhabib, line 46: should this not be 0.18?

PR #439, opened 2025-01-03
Provide num_subnet_nodes to bitcoin ckbtc_kyt

dskloetd:
# Motivation

[Updating the IC commit](https://github.com/dfinity/snsdemo/pull/437) [failed](https://github.com/dfinity/snsdemo/actions/runs/12541516365/job/34970180020?pr=437) with
```
Error: Failed while trying to deploy canisters.
Caused by: Failed while trying to install all canisters.
Caused by: Failed to install wasm module to canister 'ckbtc_kyt'.
Caused by: Failed to create argument blob.
Caused by: Invalid data: Unable to serialize Candid values: Invalid data: Unable to serialize Candid values: record field num_subnet_nodes not found
```
because https://github.com/dfinity/ic/pull/3056 added a new required field in the init arguments of the bitcoin checker.

# Changes

1. Update the pinned IC commit hash.
2. Provide `num_subnet_nodes` in the install args of the KYT/checker canister.

# Tested

Creating a snapshot fails after updating the IC commit hash but works again after specifying the `num_subnet_nodes`.

### DIFF
--- a/bin/dfx-ckbtc-deploy
+++ b/bin/dfx-ckbtc-deploy
@@ -58,7 +58,7 @@ deploy_ckbtc() {
   InitArg = record {
       btc_network = variant { mainnet };
       check_mode = variant { AcceptAll };
-      num_subnet_nodes = opt 1;
+      num_subnet_nodes = 1;
   }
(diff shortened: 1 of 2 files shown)


yhabib (approved):
Thanks for fixing this one 🙌

PR #443, opened 2025-01-20
Fail if no new IC commit

dskloetd:
# Motivation

It's very unlikely there will not be a new commit on the IC repo for a week.
So if we try to update the IC commit in snsdemo and there is no newer commit that supports all the necessary downloads, it's probably because one of the downloads moved or is broken and we should look into it.

# Changes

1. When finding the latest IC commit, output which downloads are not available for skipped IC commits.
2. When there is no new IC commit, fail the workflow so that we get a message on Slack.

# Tested

Ran [here](https://github.com/dfinity/snsdemo/actions/runs/12865338979/job/35865653929), resulting in [this slack message](https://dfinity.slack.com/archives/C05G2RZFJAJ/p1737366787388989).


yhabib (approved):
Thanks

PR #447, opened 2025-02-03
Remove saving state from demo workflow

dskloetd:
# Motivation

The "non-interactive demo" [workflow](https://github.com/dfinity/snsdemo/blob/fbfc3154966dcb70375949655cce4e132f358333/.github/workflows/run.yml#L73) broke because it used `actions/upload-artifact: v3`, which is no longer supported, to upload a saved state.

But there is no reason for the demo workflow to upload a saved state in the first place. We already have the "snapshot" [workflow](https://github.com/dfinity/snsdemo/blob/main/.github/workflows/snapshot.yml), which creates a snapshot and includes it in a release.

# Change

Stop saving and uploading the state in the demo workflow.

# Tested

CI should pass again.

### DIFF
--- a/.github/workflows/run.yml
+++ b/.github/workflows/run.yml
@@ -52,29 +52,6 @@ jobs:
         if: matrix.os != 'macos-14'
         run: |
           ./bin/dfx-sns-demo-mksns-parallel -s 10 --config_index_offset 2 -m snsdemo8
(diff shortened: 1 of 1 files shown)


mstrasinskis (approved):
Thanks!

PR #449, opened 2025-02-03
Use webserver-port instead of replica-port

dskloetd:
# Motivation

It seems to be that when running a local replica with `dfx` you can reach the webserver on the replica port (which you get with `dfx info replica-port`).
But when using PocketIC (with `dfx start --pocketic`) you need to use the webserver port (which you get with `dfx info webserver-port`) explicitly.

So in preparation to using dfx with PocketIC, we should switch to using `webserver-port`.

# Changes

1. Use `dfx info webserver-port` instead of `dfx info replica-port` when we are trying to connect to the webserver.

# Tested

1. CI still passes
2. When trying `--pocketic` locally, this change was necessary to make it work.


yhabib (approved):
Thanks

PR #450, opened 2025-02-05
Update dfx extensions to 0.5.0

dskloetd:
# Motivation

We want to migrate to using `dfx` with PocketIC.
We need a newer version of the `dfx nns` extension for `dfx nns install` to work with PocketIC and for SNSes to be installed on the SNS subnet.
The latter make it so that SNS canisters need to use cycles, which allows us to create an environment where we can test with frozen SNS canisters.

We also update the `dfx sns` extension just to keep the versions in sync.

This PR does not yet move to PocketIC.

# Changes

1. Update pinned version of nns and sns extension to 0.5.0


mstrasinskis (approved):
Thanks!

PR #451, opened 2025-02-05
Install pinned version of nns extension

dskloetd:
# Motivation

We need the newest version of the nns extension to be able to freeze SNSes.
Currently `dfx-nns-install` makes sure the extension is installed, but if it's already installed it leaves it at the current version.

# Changes

1. Make sure the pinned version of the nns extension is installed by passing `--version pinned` instead of `--if-not-installed`.

# Tested

1. Manually, freezing SNS canisters didn't work because I still had an older version of the extension. After this change I saw it correctly update the extension and I was able to freeze SNS canisters.


yhabib (approved):
Thanks

PR #455, opened 2025-02-19
Update dfx to version 0.25.0

dskloetd:
# Motivation

This is normally done periodically by an automatic workflow, but I want to move forward with using PocketIC in snsdemo and 0.25.0 is not marked "latest" yet, so I'm doing the update manually.

# Change

Use `dfx 0.25.0` in snsdemo.

# Tests

Tested manually with `--pocketic`, which will be added in a follow-up PR.


yhabib (approved):
Yeeees

PR #456, opened 2025-02-19
Add snsdemo8 as controller of SNS canisters

dskloetd:
# Motivation

Now that snsdemo uses PocketIC, we can use `--impersonate` to do things which we couldn't do before because we were not the controller of the right canisters.
To make sure this works, and to avoid having to use `--impersonate` when dealing with SNS canisters, we add `snsdemo8` as a controller to all of the SNS canisters.

# Changes

1. Add `bin/dfx-add-controller` which is copied from `scripts/add-controller` in nns-dapp.
2. Add `--add_controller` flags to `bin/dfx-sns-demo-mksns` and `bin/dfx-sns-demo-mksns-parallel`.
3. Pass `--add_controller` from `bin/dfx-sns-demo` and `bin/dfx-stock-deploy`.

# Tested

1. Created a snapshot and ran it in nns-dapp. Checked that my principal was present as a controller of various SNS canisters of the A, B and C SNSes.
2. Tested that I was able to do `dfx canister status` on those canisters.


yhabib (approved):
Nice. This simplifies the cycle depletion process for local development. 🙌

PR #458, opened 2025-02-26
Check for pocket-ic process

dskloetd:
# Motivation

In some places we check if a replica is running. Now that we switched to PocketIC, we should check if `pocket-ic` is running.

# Changes

Change `pgrep -x replica` to `pgrep -x pocket-ic`

# Tested

Tried to run `bin/dfx-snapshot-stock-make` while creating a snapshot and saw the `A replica is already running.` warning. While previously it just killed the other snapshot.


yhabib (approved):
Thanks

PR #459, opened 2025-02-26
Make pocket-ic optional

dskloetd:
# Motivation

When using a PocketIC snapshot on nns-dapp, e2e tests became flaky.
Especially the screenshot tests require a lot more time to settle before we can take a screenshot.
Until this can be fixed we shouldn't use PocketIC on nns-dapp CI yet.

# Changes

1. Add `--pocketic` flag which controls if the snapshot is made with `dfx start --pocketic`.
2. Don't try to add controllers to SNS canisters when not using PocketIC.
3. Wait for a checkpoint to be made if not running with PocketIC.
4. When checking if a snapshot is running, use both `pgrep -x replica` and `pgrep -x pocket-ic`.
5. Use `--pocketic` in `run.yml` to have some test coverage with PocketIC.

# Tested

1. Created snapshots with and without `--pocket`.
2. Used both snapshots.
3. Verified that they were with/without pocket-ic by running `jq .type ~/Library/Application\ Support/org.dfinity.dfx/network/local/replica-effective-config.json`.

### DIFF
--- a/.github/workflows/run.yml
+++ b/.github/workflows/run.yml
@@ -42,7 +42,7 @@ jobs:
       - name: Install dependencies
         run: ./bin/dfx-sns-demo-install --verbose
       - name: Run the demo with the current default ic commits
-        run: bin/dfx-sns-demo --config_index 0
+        run: bin/dfx-sns-demo --config_index 0 --pocketic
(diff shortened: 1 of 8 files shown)


yhabib (approved):
Thanks

PR #460, opened 2025-02-26
chore(deps): update nns-dapp pin to latest proposal version

yhabib:
# Motivation

`dfx nns install` installs an old version of nns-dapp. Since then some changes affecting its stable memory have been made.

# Changes

- Update the pinned release of nns-dapp to the newest proposal.

# Tested

- Manually created a snapshot and used it with nns-dapp.


mstrasinskis (approved):
Thank you!

